PR #11, opened 2020-10-24
Reformat code

agluszak:
Reformat code using google-java-format for Java (some files were already formatted using it) and Bazel plugin for .bzl files


andrefmrocha:
Thanks for this! Is there any way we can enforce it with a linter?

agluszak:
I can hack together a git pre-commit hook

PR #12, opened 2020-10-25
Change project structure

agluszak:
- Separated the code into 3 directories: Server, Install and Common with their own BUILD files
- Added proper CLI to the installer (with an option to use the debugger)
- Moved more "magic strings" to Constants file
- Simplified Server.java (removed the option to launch only BEP server as it was unused)

Tests may be broken for now, @Abrams27 will soon submit a refactoring of the server code and then we can fix them.


.bazelrc
```diff
@@ -1,2 +1,2 @@
-build --java_toolchain=@bazel_tools//tools/jdk:toolchain_java8
-build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java8
\ No newline at end of file
+build --java_toolchain=@bazel_tools//tools/jdk:toolchain_java9
```
Abrams27, line 1: why not 11?

main/src/org/jetbrains/bsp/bazel/common/Constants.java
```diff
@@ -0,0 +1,19 @@
+package org.jetbrains.bsp.bazel.common;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+public class Constants {
```
Abrams27, line 6: do we want to keep it in the class or move it to the `.properties` file? https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html

main/src/org/jetbrains/bsp/bazel/install/Install.java
```diff
@@ -0,0 +1,214 @@
...
+        Option.builder(DEBUGGER_SHORT_OPT)
+            .longOpt("debugger")
+            .hasArg()
+            .argName("address (e.g. '127.0.0.1:8000'")
+            .desc("Allow BSP server debugging")
+            .build();
+
+    cliOptions.addOption(debug);
+
+    Option directory =
```
Abrams27, line 177: It looks veeery similar to the `java` and `bazel` initialization. Can we extract it to the separate method?

main/src/org/jetbrains/bsp/bazel/install/Install.java
```diff
@@ -0,0 +1,214 @@
...
+    HelpFormatter formatter = new HelpFormatter();
+    formatter.setWidth(120);
+    boolean hasError = false;
+    PrintWriter writer = new PrintWriter(System.err);
+    try {
+      CommandLine cmd = parser.parse(cliOptions, args, false);
+      if (cmd.hasOption(HELP_SHORT_OPT)) {
+        formatter.printHelp(INSTALLER_BINARY_NAME, cliOptions);
+      } else {
+        // Add Java binary
```
Abrams27, line 53: i think that extracting this section to the method `addJavaBinary` would help to remove this comment.

main/src/org/jetbrains/bsp/bazel/install/Install.java
```diff
@@ -0,0 +1,214 @@
...
+        if (cmd.hasOption(JAVA_SHORT_OPT)) {
+          argv.add(cmd.getOptionValue(JAVA_SHORT_OPT));
+        } else {
+          String javaHome = readSystemProperty("java.home");
+          argv.add(Paths.get(javaHome).resolve("bin").resolve("java").toString());
+        }
+        // Add Java classpath
```
Abrams27, line 60: as above

main/src/org/jetbrains/bsp/bazel/install/Install.java
```diff
@@ -0,0 +1,214 @@
...
+        argv.add("-classpath");
+        String javaClassPath = readSystemProperty("java.class.path");
+        Splitter.on(":").splitToList(javaClassPath).stream()
+            .map(elem -> Paths.get(elem).toAbsolutePath().toString())
+            .forEach(argv::add);
+        // Add debugger connection
```
Abrams27, line 66: as above

main/src/org/jetbrains/bsp/bazel/install/Install.java
```diff
@@ -0,0 +1,214 @@
...
+        if (cmd.hasOption(DEBUGGER_SHORT_OPT)) {
+          String debuggerAddress = cmd.getOptionValue(DEBUGGER_SHORT_OPT);
+          argv.add(
+              "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=" + debuggerAddress);
+        }
+        argv.add(SERVER_CLASS_NAME);
+        // Add Bazel binary
```
Abrams27, line 73: as above

main/src/org/jetbrains/bsp/bazel/install/Install.java
```diff
@@ -0,0 +1,214 @@
...
+      System.exit(1);
+    }
+  }
+
+  private static void writeConfigurationFiles(Path rootDir, Object discoveryDetails)
+      throws IOException {
+    Path bazelbspDir = rootDir.resolve(Constants.BAZELBSP_DIR_NAME);
+    Files.createDirectories(bazelbspDir);
+
+    // Copy aspects.bzl file
```
Abrams27, line 118: as above

main/src/org/jetbrains/bsp/bazel/install/Install.java
```diff
@@ -0,0 +1,214 @@
...
+    Path aspectsFile = bazelbspDir.resolve(Constants.ASPECTS_FILE_NAME);
+    Files.copy(
+        Install.class.getResourceAsStream(Constants.ASPECTS_FILE_NAME),
+        aspectsFile,
+        StandardCopyOption.REPLACE_EXISTING);
+
+    // Create an empty BUILD file, overwrite if exists
```
Abrams27, line 125: as above

Abrams27 (commented):
I'm not a fan of such "script" code in java even though it is some kind of a script. Nevertheless, good job, looks amazing!

test.sh
```diff
@@ -8,7 +8,7 @@ test_bsp_server() {
   bazel run --define "maven_repo=file://$HOME/.m2/repository" //:bsp.publish
   bsp_path="$(bazel info bazel-bin)/bsp-project.jar"
   cd sample-repo
-  java -cp $bsp_path org.jetbrains.bsp.bazel.Install
+  java -cp $bsp_path org.jetbrains.bsp.bazel.install.Install
```
andrefmrocha, line 11: This change also needs to be reflected in the Readme. Can you change that?

andrefmrocha (changes requested):
Thanks for this! It does reflect some very much needed structural changes :smiley: . 
Just pointed out some minor nitpickings that need to be taken into account

.bazelrc
agluszak: https://www.baeldung.com/java-9-process-api#triggering-dependent-actions-on-process-termination

agluszak: (we don't use it yet, but we discussed with @Abrams27 that we'd like to change some async logic in the server and this would allow for some improvements, but of course this method is not indispensable)

andrefmrocha: Ah, I was not aware Java9 had introduced such an API, I agree with its use, not sure if we can abdicate the use of java8 just like that, I'd advise to discuss this with @jastice

Abrams27: if only we could update java version then we should use 11 (LTS) instead of 9

main/src/org/jetbrains/bsp/bazel/common/Constants.java
agluszak: Maybe I'll leave it as it is for now, reading properties will require a lot of parsing and/or typecasting and it won't provide any value for us, because these constants are non-configurable anyway

main/src/org/jetbrains/bsp/bazel/install/Install.java
agluszak: There's no need to, calling the extracted method would look almost the same - we still need to pass the arguments.

test.sh
agluszak: I didn't change it on purpose: Readme describes installation process with coursier, but we didn't publish a new version to Maven, so that instruction is still valid.

main/src/org/jetbrains/bsp/bazel/common/Constants.java
Abrams27: okey, but I'll change it in the future

Abrams27 (approved):
looks good

test.sh
andrefmrocha: The instructions for coursier actually mention the snapshots repository of Maven, that Teamcity will automatically push to after a merge to master.
Perhaps you don't remember from the deep-dive we did (I wouldn't remember either, from the amount of information I dropped on you :smile: ), but there's a CI/CD running on this that we cannot see on Github

PR #13, opened 2020-11-03
[1 - Initial code refactor] basic classes extraction

Abrams27:
First step of the initial code refactor.

Introducing:
- `ScalaBuildServer` implementation moved to the separate class
- `JavaBuildServer` implementation moved to the separate class
- common methods calls for `ScalaBuildServer` and `JavaBuildServer` moved to the separate resolvers
- changed logging
- changed a bit initialization in the `Install`
- added obligatory brackets in the if statements

Only methods bodies have been moved to separate classes, I didn't extract new methods nor did I change logic.

Initial code refactor roadmap:
1) basic responsibilities classes extraction < this PR
2) adding unit tests for extracted classes
3) extracting more logic to separate classes (OOP SOLID implementation)
4) more unit tests
5) integration tests (?)


main/src/org/jetbrains/bsp/bazel/BUILD
```diff
@@ -1,6 +1,6 @@
 java_library(
     name = "bsp",
-    srcs = glob(["*.java"]),
+    srcs = glob(["**/*.java"]),
```
agluszak, line 3: This is not a good change, but my pr (#12) will take care of it

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
```diff
@@ -111,88 +109,46 @@
...
-              "Server has not been initialized yet!",
-              false));
-    if (isFinished())
-      return completeExceptionally(
-          new ResponseError(
-              ResponseErrorCode.serverErrorEnd, "Server has already shutdown!", false));
-
-    return getValue(request);
+  public void setBackendPort(int port) {
+    this.BES_BACKEND += port;
```
agluszak, line 147: BES_BACKEND looks like a constant, but it's not. And what if you call setBackendPort twice?

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
```diff
@@ -859,7 +741,7 @@ private String convertOutputToPath(String output, String prefix) {
 
             int returnCode = parseProcess(process);
             return Either.forRight(new TestResult(BepServer.convertExitCode(returnCode)));
-          } catch (IOException | InterruptedException | ExecutionException e) {
+          } catch (IOException | InterruptedException e) {
```
agluszak, line 744: why was ExecutionException removed?

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
```diff
@@ -888,20 +773,40 @@ private String convertOutputToPath(String output, String prefix) {
 
             int returnCode = parseProcess(process);
             return Either.forRight(new RunResult(BepServer.convertExitCode(returnCode)));
-          } catch (IOException | InterruptedException | ExecutionException e) {
+          } catch (IOException | InterruptedException e) {
```
agluszak, line 776: Same as above

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
```diff
@@ -912,156 +817,104 @@ private String convertOutputToPath(String output, String prefix) {
...
-                          target ->
-                              collectScalacOptionsResult(
-                                  either.getRight(),
-                                  targetsOptions.getOrDefault(target, new ArrayList<>()),
-                                  either.getRight().getInputsAsUri(target, getExecRoot()),
-                                  target))
-                      .collect(Collectors.toList()));
-          return Either.forRight(result);
-        });
+    return executeCommand(() -> scalaBspServer.buildTargetScalacOptions(scalacOptionsParams));
```
agluszak, line 820: Maybe this whole method should be moved to ScalaBspServer? I know that currently this is done how it is because BazelBspServer acts as a "hub" for all json APIs, but maybe it doesn't have to be that way?

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
```diff
@@ -912,156 +817,104 @@ private String convertOutputToPath(String output, String prefix) {
...
   @Override
   public CompletableFuture<JavacOptionsResult> buildTargetJavacOptions(
       JavacOptionsParams javacOptionsParams) {
-    return executeCommand(
-        () -> {
-          List<String> targets =
-              javacOptionsParams.getTargets().stream()
-                  .map(BuildTargetIdentifier::getUri)
-                  .collect(Collectors.toList());
+    return executeCommand(() -> javaBspServer.buildTargetJavacOptions(javacOptionsParams));
```
agluszak, line 826: same as above

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
```diff
@@ -912,156 +817,104 @@ private String convertOutputToPath(String output, String prefix) {
...
+  }
 
-          String targetsUnion = Joiner.on(" + ").join(targets);
-          Map<String, List<String>> targetsOptions = getTargetsOptions(targetsUnion, "javacopts");
-          // TODO(andrefmrocha): Remove this when kotlin is natively supported
-          Either<ResponseError, ActionGraphParser> either =
-              parseActionGraph(getMnemonics(targetsUnion, Lists.newArrayList(JAVAC, KOTLINC)));
-          if (either.isLeft()) return Either.forLeft(either.getLeft());
+  @Override
```
agluszak, line 829: same as above

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
```diff
@@ -912,156 +817,104 @@ private String convertOutputToPath(String output, String prefix) {
...
-                              collectJavacOptionsResult(
-                                  either.getRight(),
-                                  targetsOptions.getOrDefault(target, new ArrayList<>()),
-                                  either.getRight().getInputsAsUri(target, getExecRoot()),
-                                  target))
-                      .collect(Collectors.toList()));
-          return Either.forRight(result);
-        });
+  @Override
+  public CompletableFuture<ScalaMainClassesResult> buildTargetScalaMainClasses(
```
agluszak, line 836: same as above

main/src/org/jetbrains/bsp/bazel/resolvers/QueryResolver.java
```diff
@@ -0,0 +1,28 @@
+package org.jetbrains.bsp.bazel.resolvers;
+
+import com.google.devtools.build.lib.query2.proto.proto2api.Build;
+import java.io.IOException;
+import java.util.concurrent.Semaphore;
+
+public class QueryResolver {
+
+  private final ProcessResolver processResolver;
+  private final Semaphore processLock;
```
agluszak, line 10: so now we have more than one semaphore... are you sure that it works the same way as before?

main/src/org/jetbrains/bsp/bazel/servers/JavaBspServer.java
```diff
@@ -0,0 +1,87 @@
...
+import org.jetbrains.bsp.bazel.ActionGraphParser;
+import org.jetbrains.bsp.bazel.Uri;
+import org.jetbrains.bsp.bazel.resolvers.ActionGraphResolver;
+import org.jetbrains.bsp.bazel.resolvers.TargetsResolver;
+import org.jetbrains.bsp.bazel.utils.MnemonicsUtils;
+
+// TODO: This class *should* implement a `JavaBuildServer` interface,
+// TODO: now `buildTargetJavacOptions` method returns a `Either<ResponseError, JavacOptionsResult>`
+// TODO: instead of a `CompletableFuture<JavacOptionsResult>` because of the `BazelBspServer`
+// TODO: command executing (`executeCommand`) implementation.
```
agluszak, line 25: Ahh, okay, this explains my question from above :P

main/src/org/jetbrains/bsp/bazel/servers/JavaBspServer.java
```diff
@@ -0,0 +1,87 @@
...
+public class JavaBspServer {
+
+  private final TargetsResolver targetsResolver;
+  private final ActionGraphResolver actionGraphResolver;
+
+  private final String javac;
+  private final String kotlinc;
+  private final String execRoot;
+
+  // TODO: too many arguments!!, constants will be moved
```
agluszak, line 35: constants - do you mean javac and kotlinc?

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
Abrams27: I didn't change the logic, I only moved this method. Nevertheless I agree, it should be changed.

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
Abrams27: Because Semaphore doesn't throw it (`CompletableFuture` does).

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
Abrams27: I think that there was a reason for it, we can discuss it later.

main/src/org/jetbrains/bsp/bazel/servers/JavaBspServer.java
Abrams27: yes

main/src/org/jetbrains/bsp/bazel/resolvers/ProcessResolver.java
```diff
@@ -0,0 +1,79 @@
...
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.Semaphore;
+
+public class ProcessResolver {
+
+  private final Semaphore processLock;
```
andrefmrocha, line 16: Just a minor note: this is a great solution for preventing any process concurrency issues but I was able to find that these weren't in any way a real issue, since all requests are done sequentially.

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
andrefmrocha: Might I get updated on what was the reason you guys were thinking of? Since it apparently was moved either way

main/src/org/jetbrains/bsp/bazel/resolvers/ProcessResolver.java
Abrams27: There was a variable `processLock` which was used as ... a lock! (more precisely as a semaphore) I just changed it to the more elegant solution. Does it mean that it is redundant and even original `processLock` was redundant?

andrefmrocha: As of that time yes it was, but since I hadn't tested this against other BSP implementations rather the IntelliJ one, I thought it wouldn't be a good idea to make assumptions

jastice: In the general case we should assume that requests by a client aren't necessarily sequential, even if only for the sake of the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle)

andrefmrocha: But since that isn't something we can take as guaranteed from clients (a lot of requests can be done in parallel), this is very much welcome, I like that you guys are doing so many changes I had in mind in terms of refactors :smile:

Abrams27: Nice to hear this ☺️

main/src/org/jetbrains/bsp/bazel/BazelBspServer.java
agluszak: @andrefmrocha 
`public class BazelBspServer implements BuildServer, ScalaBuildServer, JavaBuildServer {`
XBuildServer interfaces are responsible for JSON communication with the outside world, right? What I meant was that it would be nicer if there were separate classes, each implementing one interface

andrefmrocha: Yeah I got that, but I thought @Abrams27 was mentioning there was something stopping us from implementing it that?

Abrams27: I mentioned that there is a problem in the actual implementation with `BazelBspServer` "HUB". It can be moved to separate classes and then problem would be solved.

Abrams27: I can do it in the following PR

andrefmrocha: Sure! Sounds great, thanks for keeping me up to date

src/main/java/org/jetbrains/bsp/bazel/server/logger/BuildClientLogger.java
```diff
@@ -0,0 +1,25 @@
...
+  private final BuildClient buildClient;
+
+  public BuildClientLogger(BuildClient buildClient) {
+    this.buildClient = buildClient;
+  }
+
+  public void logError(String errorMessage) {
+    LogMessageParams params = new LogMessageParams(MessageType.ERROR, errorMessage);
+    buildClient.onBuildLogMessage(params);
+    throw new RuntimeException(errorMessage);
```
agluszak, line 18: Throwing an error inside a logging method looks suspicious (inb4 I know it used to be so even before the refactoring)

Abrams27: added TODO

PR #15, opened 2020-11-14
[configuration] Optional argument on server startup

ex0ns:
This is a draft that aims to add an optional parameters when starting a server.
The second argument must be a valid bazel's package. Bazel query capabilities are then used to find all the targets for this specific package.

Client can now spawn a server with the following syntax:

```sh
java -classpath "$HOME/.m2/repository/org/jetbrains/bsp/bazel-bsp/0.1.0/bazel-bsp-0.1.0.jar" org.jetbrains.bsp.bazel.Server /usr/local/bin/bazel //toy
```

To reduce the scope of the package to the one named `toy`.
This is not a definitive version, it's only here to open discussion about this topic. I think we will need more advanced argument parsing for that (first reason is that bazel's path is optional and hence the position of arguments might differs).

Any suggestion/feedback welcome (I'm not used to Java anymore, review and advice would be awesome on that as well)


main/src/org/jetbrains/bsp/bazel/BazelBspServerConfig.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public BazelBspServerConfig(String pathToBazel) {
+        this.pathToBazel = pathToBazel;
+    }
+
+    static public BazelBspServerConfig from(String[] args) {
+        if(args.length == 0) { throw new IllegalArgumentException("Configuration can't be built without any parameters"); }
+        BazelBspServerConfig config = new BazelBspServerConfig(args[0]);
+        if(args.length == 2) {
+            config.setProjectPath(args[1]);
```
andrefmrocha, line 17: I think this may be too limitating for the usages of bazel. For instance we might be using the following structure:
- src
  * proj1
  * proj2
  * proj3
And we may want to import proj1 and proj3 but not proj1. This is a simple example but that can become much more complicated

main/src/org/jetbrains/bsp/bazel/BazelBspServerConfig.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public String getProjectPath() {
+        return this.projectPath;
+    }
+
+    public BazelBspServerConfig setProjectPath(String projectPath) {
+        this.projectPath = projectPath;
+        return this;
+    }
+}
```
agluszak, line 35: No newline

agluszak (commented):
Thanks for your PR! Could you change the target branch to dev from master and rebase?

main/src/org/jetbrains/bsp/bazel/BazelBspServerConfig.java
ex0ns: That is a really interesting point (I guess you meat `but not proj2`) , but in this specific case, would you want to import proj1 and proj3 because there is a dependency between the two ? Or it because you would like to work on them at the same time ? (I rarely open my whole monorepo into a single editor, I tend to open project by project and spawn a different server for each one of them, this is why I'm asking).
Assuming we want to keep this approach:
1. Should we provide the ability to the user to load a list of target instead of a single one ?
2. Should we do something about source dependencies ?

andrefmrocha: Let's say that it is because you want to work on both. I didn't think this would be an issue either but after some talk with people who frequently work in monorepo structures, we foud ou that this was a common necessity.
Regarding your questions:
1. I think having the ability to select the targets would be wonderful
2. Source dependencies will have to be indexed either way but that shouldn't require any more work into it after specifiying the dependencies, at least I don't think so

ex0ns:
> Thanks for your PR! Could you change the target branch to dev from master and rebase?

Done (`Server.java` is much cleaner now).
I did not add the support for multiple targets yet, I would like to have some bazel advice first:

Do we agree that we want to be able to specify the target but not a specific rules inside a target (I'm not sure about the names used in Bazel) ? 

For instance, we want to be able to specify that we want to build `//proj1` and `//proj2` but we don't want to specify `//proj1:rule1`, `//proj1:rules2` and so on ? 
In other world, do we keep the query

```
kind(binary, %s:all) union kind(library, %s:all) union kind(test, %s:all)
```

For each target ? Or do we want something more fine grained ?

andrefmrocha:
> > Thanks for your PR! Could you change the target branch to dev from master and rebase?
> 
> Done (`Server.java` is much cleaner now).
> I did not add the support for multiple targets yet, I would like to have some bazel advice first:
> 
> Do we agree that we want to be able to specify the target but not a specific rules inside a target (I'm not sure about the names used in Bazel) ?
> 
> For instance, we want to be able to specify that we want to build `//proj1` and `//proj2` but we don't want to specify `//proj1:rule1`, `//proj1:rules2` and so on ?
> In other world, do we keep the query
> 
> ```
> kind(binary, %s:all) union kind(library, %s:all) union kind(test, %s:all)
> ```
> 
> For each target ? Or do we want something more fine grained ?

I think it's fine for us to work on that level yeah! Don't have that much experience with monorepos, but it's hard to see a use case where that is relevant

main/src/org/jetbrains/bsp/bazel/BazelBspServerConfig.java
ex0ns: It is now possible to use multiple targets when starting the server by specifying them using a comma as delimiter:

```sh 
java -classpath bazel-bsp-0.1.0-SNAPSHOT.jar org.jetbrains.bsp.bazel.server.Server /usr/local/bin/bazel //toy2,//toy,//lib/otherlib
```

Is there something else we should take into account ?

Abrams27 (commented):
Great job! I added three minor comments.

src/main/java/org/jetbrains/bsp/bazel/server/Server.java
```diff
@@ -49,7 +47,8 @@ public static void main(String[] args) {
       System.setOut(logStream);
       System.setErr(logStream);
 
-      BazelBspServer bspServer = new BazelBspServer(bazelPath);
+      BazelBspServerConfig config = BazelBspServerConfig.from(args);
```
andrefmrocha, line 50: In order to not have to add this values manually to the json file created with the startup options, we we probably should put these ourselves in the json file. Wdyt?

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServerConfig.java
```diff
@@ -0,0 +1,34 @@
+package org.jetbrains.bsp.bazel.server;
+
+public class BazelBspServerConfig {
+
+    private String pathToBazel = null;
+    private String[] targetProjectPaths = new String[] { "//..." };
```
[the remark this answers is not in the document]
ex0ns: I can, but I used Array on purpose.
The idea behind it is that this configuration is supposed to be "static" once the server started, there should be no reason to add/remove target (in the configuration at least). I know that Array won't prevent that but their semantic is pretty different from list (it's also the reason why I did not add any `addTarget` or `removeTarget` on the configuration).
What do you think ?

src/main/java/org/jetbrains/bsp/bazel/server/Server.java
ex0ns: I'm not sure to understand, are you talking about the json generated by `org.jetbrains.bsp.bazel.Install` ?
Do you think that we should add to `argv` the target of the current folder ? I guess that could be done (either via a bazel query of by finding the path up to a WORKSPACE file)

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServerConfig.java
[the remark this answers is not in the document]
jastice: Sounds more like a case for an immutable / unmodifiable list :)

src/main/java/org/jetbrains/bsp/bazel/server/Server.java
jastice: I think it means we could pass the desired targets to the `Install` arguments and in turn pass them into `argv`

jastice:
One use-case to consider: Before importing, we might want to query the available targets in the workspace, and let the user select the ones of interest in the IDE, given some Bazel-specific support. I prsume this would be handled by a regular Bazel query, and then appending these targets to the argv?

src/main/java/org/jetbrains/bsp/bazel/server/Server.java
andrefmrocha: Yeah, as @jastice said :)

main/src/org/jetbrains/bsp/bazel/BazelBspServerConfig.java
andrefmrocha: maybe we should also be able to receive separated by a space, wdyt?

andrefmrocha:
> One use-case to consider: Before importing, we might want to query the available targets in the workspace, and let the user select the ones of interest in the IDE, given some Bazel-specific support. I prsume this would be handled by a regular Bazel query, and then appending these targets to the argv?

Yeah that can be a way of doing it, or maybe we can use the the `workspaceBuildTargets` requests to obtain the valid workspaces beforehand and then reinstall the server with it :thinking:

src/main/java/org/jetbrains/bsp/bazel/server/Server.java
ex0ns: I've just added the `--targets` flag on the `Install`

ex0ns:
The issue with the `workspaceBuildTargets` (as I see it now) it that according to the documentation, it should be able to take an argument (https://build-server-protocol.github.io/docs/specification#workspace-build-targets-request) however this seems that the java implementation does not provide this parameters. @andrefmrocha did I miss something ? I don't know how the client and server could communicate to update the `buildTargets`

andrefmrocha:
> The issue with the `workspaceBuildTargets` (as I see it now) it that according to the documentation, it should be able to take an argument (https://build-server-protocol.github.io/docs/specification#workspace-build-targets-request) however this seems that the java implementation does not provide this parameters. @andrefmrocha did I miss something ? I don't know how the client and server could communicate to update the `buildTargets`

But the workspaceBuildTargets can be used to obtain all of the targets of the project so that the client provides a means for the user to select the ones he wants. Not sure where you got that this requests can receive an argument, the interface for the request is empty. Perhaps you were looking at the response?

src/main/java/org/jetbrains/bsp/bazel/server/Server.java
andrefmrocha: Looks great! Thanks!

main/src/org/jetbrains/bsp/bazel/BazelBspServerConfig.java
ex0ns: I can implement it if you think that is a nice to have but I'm not a really big fan of that.
What we could support on the other hand is having (like in the Install) a real command line argument parsing in the server, this could allow us to have multiple `--target (-t)` targets flags when starting the server 

```
org.jetbrains.bsp.bazel.server.Server /usr/local/bin/bazel -t //toy2 -t //toy -t //lib/otherlib
```

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServerConfig.java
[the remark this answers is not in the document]
ex0ns: Update the code. Coming from the Scala world, I don't know if my implementation is the "state of the art" in Java, please tell me if I need to change something.

ex0ns:
I think I get it, so so workflow would basically be:

1. Start the server 
2. Get the list of targets from the server in the client
3. Update the server's configuration
4. Restart the server with the new configuration (with the selected targets)

My question is: what is happening when the user selects the targets they want in the client ? How do we communicate the result back to the server in order to update the configuration

andrefmrocha:
> I think I get it, so so workflow would basically be:
> 
> 1. Start the server
> 2. Get the list of targets from the server in the client
> 3. Update the server's configuration
> 4. Restart the server with the new configuration (with the selected targets)
> 
> My question is: what is happening when the user selects the targets they want in the client ? How do we communicate the result back to the server in order to update the configuration

I think that behaviour would be implemented in the client side, without any sort of server support. Or if there was, an extension to the protocol would have to be made

ex0ns:
I add a busy week, I'm currently trying to make the tests pass.
I noticed that `workspaceBuildTargets` and I still don't know why, I'll update this PR when fixed.

```
[Trace - 06:56:46 PM] Sending response 'workspace/buildTargets - (2)'. Processing request took 31940ms
Result: {
  "items": []
}
```

andrefmrocha (approved):
Lgtm! :shipit:

agluszak:
Hi @ex0ns! Thanks again for your contribution. I made a PR to your fork https://github.com/ex0ns/bazel-bsp/pull/1/files :) I thought it would be easier to do this instead of another round of review as this is pure nitpicking. When you merge my PR in your repo I'll merge yours here

Google style guide prohibits wildcard imports - but intellij introduces them by default, even when you install google-java-format plugin. You can disable them here https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#exclude-from-auto-import

PR #18, opened 2020-11-22
Refactor pt 2

agluszak:
- Introduced `BazelRunner` class which is fully responsible for starting processes, awaiting them and gathering results
- Moved synchronized `get<BazelData>` methods to `BazelDataResolver`. They are no longer required to be synchronized, as all the required data is fetched in the constructor of `BazelBspServer`
- Removed checked exceptions in many places and the need for try-to-rethrow
- Removed `Either`s in many places where they were no longer needed
- Corrected several minor issues (`setBackendPort` appending to a constant, `convertExitCode` not using enum names)


Abrams27 (commented):
Great job! I added 3 minor comments - later, I will look one more time more closely.

src/main/java/org/jetbrains/bsp/bazel/server/resolvers/QueryResolver.java
```diff
@@ -2,26 +2,21 @@
...
 import com.google.devtools.build.lib.query2.proto.proto2api.Build;
 import java.io.IOException;
-import java.util.concurrent.Semaphore;
+import org.jetbrains.bsp.bazel.server.data.ProcessResults;
 
 public class QueryResolver {
 
-  private final ProcessResolver processResolver;
-  private final Semaphore processLock;
+  private final BazelRunner bazelRunner;
```
Abrams27, line 9: why did u remove a semaphore?

agluszak: because locking and releasing it is now a responsibility of the ProcessResolver

Abrams27: did u mean `BazelRunner`?

agluszak: yup

andrefmrocha (approved):
Looks good! Good job!

PR #21, opened 2020-12-02
BEP Server refactor

gerardd33:
This is a draft of my refactor of BepServer (_now finished_).

### What I did
- Tried to make it more readable, improvements of style, consistency, arrangement etc.
- In most places refactored, extracted methods, rewrote using streams, simplified, generally cleaned up.
- Extracted and tidied up constants (also partially in other files but still some work to be done there) 
- Extracted the inline object returned in publishBuildToolEventStream to BepStreamObserver.
- Extracted DiagnosticsDispatcher.
- Extracted ParsingUtils.
- Added a logger for src/main and replaced System.outs and errs in BepServer with it (logging itself should be improved there though).

### What I didn't do

Main points where I wasn't sure what to do, so a lot of suggestions in the review would be great:

1. _RESOLVED_

2. _RESOLVED_

3. _RESOLVED_

### TODO

_RESOLVED_


src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
-                .build();
-        responseObserver.onNext(response);
+    return new BepStreamObserver(this, responseObserver);
+  }
+
+  public TreeSet<Uri> getCompilerClasspath() {
+    return compilerClasspath;
+  }
+
+  public Map<String, String> getDiagnosticsProtosLocations() {
```
Abrams27, line 97: I think that getter should be moved to the bottom of the class

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
+    return diagnosticsProtosLocations;
+  }
+
+  public void collectDiagnostics(
```
Abrams27, line 101: I believe that this function should return this collected Diagnostics, input argument (`files ToDiagnostics`) isn't the best idea.

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
+    diagnosticsDispatcher.collectDiagnostics(filesToDiagnostics, target, diagnosticsLocation);
+  }
+
+  public void emitDiagnostics(
+      Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics, BuildTargetIdentifier target) {
+    diagnosticsDispatcher.emitDiagnostics(filesToDiagnostics, target);
+  }
+
+  public void handleEvent(BuildEvent buildEvent) {
+    try {
```
Abrams27, line 115: In my opinion entire logic from `try - catch` block should be moved to the separate method:
```
try {
  <method>
} ...

File: src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java [hunk 4]
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
+      BuildEventStreamProtos.BuildEvent event =
+          BuildEventStreamProtos.BuildEvent.parseFrom(buildEvent.getBazelEvent().getValue());
+
+      LOGGER.info("Got event" + event + "\nevent-end\n");
```
Abrams27, line 119: What does `"\nevent-end\n"` mean?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
+
+      if (event.hasStarted() && event.getStarted().getCommand().equals("build")) {
```
Abrams27, line 121: I think that this methods could be changed a bit:
```
if (event.hasStarted() && event.getStarted().getCommand().equals("build")) {
  processedStartedEvent(event.getStarted());
}
```

could be moved to the separate  method with correct name

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
 
-      @Override
-      public void onCompleted() {
-        responseObserver.onCompleted();
+      if (event.hasProgress()) {
+        processProgressEvent(event.getProgress());
       }
-    };
+    } catch (IOException e) {
+      LOGGER.error("Error deserializing BEP proto: " + e);
```
Abrams27, line 149: logger methods work with placeholders:
```
LOGGER.error("Error deserializing BEP proto: {}", e);
```

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
-        .forEach(
-            error -> {
-              String erroredFile = error.contains(workspace) ? workspace : build;
-              String[] lineLocation;
-              String fileLocation;
+  private void processFinishedEvent(BuildEventStreamProtos.BuildFinished buildFinished) {
+    if (taskParkingLot.empty()) {
+      LOGGER.info("No start event id was found.");
+      return;
+    } else if (taskParkingLot.size() > 1) {
```
Abrams27, line 166: `else` is unnecessary here

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
+    bspClient.onBuildTaskFinish(finishParams);
   }
 
-  private void processCompletionEvent(BuildEventStreamProtos.BuildEvent event) throws IOException {
-    List<BuildEventStreamProtos.OutputGroup> outputGroups =
-        event.getCompleted().getOutputGroupList();
+  private void processProgressEvent(BuildEventStreamProtos.Progress progress) {
+    HashMap<String, List<Diagnostic>> fileDiagnostics = parseStdErrDiagnostics(progress);
+
+    for (Entry<String, List<Diagnostic>> entry : fileDiagnostics.entrySet()) {
```
Abrams27, line 181: it can be done im a more functional way

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +87,122 @@ public void publishLifecycleEvent(
...
+    List<OutputGroup> outputGroups = targetComplete.getOutputGroupList();
+
     if (outputGroups.size() == 1) {
       BuildEventStreamProtos.OutputGroup outputGroup = outputGroups.get(0);
-      if ("scala_compiler_classpath_files".equals(outputGroup.getName())) {
+
+      if (outputGroup.getName().equals("scala_compiler_classpath_files")) {
         for (BuildEventStreamProtos.BuildEventId.NamedSetOfFilesId fileSetId :
             outputGroup.getFileSetsList()) {
           for (BuildEventStreamProtos.File file :
```
Abrams27, line 208: nested fors should be extracted to the methods

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -266,132 +234,98 @@ private void processCompletionEvent(BuildEventStreamProtos.BuildEvent event) thr
...
       return;
     }
-    System.out.println("DWH: Event: " + event + "\n\n");
-    Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics = new HashMap<>();
+
     BuildTargetIdentifier target = new BuildTargetIdentifier(action.getLabel());
     boolean hasDiagnosticsOutput = diagnosticsProtosLocations.containsKey(target.getUri());
+    Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics = new HashMap<>();
+
     for (BuildEventStreamProtos.File log : action.getActionMetadataLogsList()) {
```
Abrams27, line 255: it can be done in a functional way too

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -266,132 +234,98 @@ private void processCompletionEvent(BuildEventStreamProtos.BuildEvent event) thr
...
-          Uri.fromExecOrWorkspacePath(
-              fileDiagnostics.getPath(),
-              bspServer.getBazelData().getExecRoot(),
-              bspServer.getBazelData().getWorkspaceRoot()),
-          convert(target, fileDiagnostics));
-    }
-  }
+    Arrays.stream(progress.getStderr().split("\n"))
+        .filter(
+            error ->
```
Abrams27, line 285: I think that this predicate can be extracted too

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -266,132 +234,98 @@ private void processCompletionEvent(BuildEventStreamProtos.BuildEvent event) thr
...
+                error.contains("ERROR")
+                    && (error.contains("/" + Constants.WORKSPACE_FILE_NAME + ":")
+                        || error.contains("/" + Constants.BUILD_FILE_NAME + ":")))
+        .forEach(
+            error -> {
```
Abrams27, line 290: as well as this huuuge lambda

src/main/java/org/jetbrains/bsp/bazel/server/BepStreamObserver.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public BepStreamObserver(
+      BepServer bepServer, StreamObserver<PublishBuildToolEventStreamResponse> responseObserver) {
+    this.bepServer = bepServer;
+    this.responseObserver = responseObserver;
+  }
+
+  @Override
+  public void onNext(PublishBuildToolEventStreamRequest request) {
+    if (request
```
Abrams27, line 24: predicate can be extracted too

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,139 @@
...
+
+  private final BazelBspServer bspServer;
+  private final BuildClient bspClient;
+
+  public DiagnosticsDispatcher(BazelBspServer bspServer, BuildClient bspClient) {
+    this.bspServer = bspServer;
+    this.bspClient = bspClient;
+  }
+
+  public void collectDiagnostics(
```
Abrams27, line 37: it is connected with previous comment - this function should return this map instead of taking it as an argument

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,139 @@
...
+    Diagnostic diagnostic = new Diagnostic(range, diagProto.getMessage());
+    if (severity != null) {
+      diagnostic.setSeverity(severity);
+    }
+
+    return diagnostic;
+  }
+
+  private DiagnosticSeverity convertSeverity(Diagnostics.Severity protoSeverity) {
+    switch (protoSeverity) {
```
Abrams27, line 120: i believe that it can be done with map instead of the `swich case `

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,139 @@
...
+      LOGGER.info("Inserting diagnostics for path: " + fileDiagnostics.getPath());
+
+      filesToDiagnostics.put(
+          getUriForPath(fileDiagnostics.getPath()), convertDiagnostics(target, fileDiagnostics));
+    }
+  }
+
+  public void emitDiagnostics(
+      Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics, BuildTargetIdentifier target) {
+    for (SourceItem source : bspServer.getCachedBuildTargetSources(target)) {
```
Abrams27, line 55: nested fors can be extracted

Abrams27 (changes requested):
Good job, few things could be changed in my opinion, but in general looks great

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: Don't know, I left the original logs, only changed the logging method. Writing proper logs (they're very fragmentary now) is definitely a task for later.

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: Thought about it but couldn't find a good name and forgot it. Do you have something better than hasBuildStarted(event)?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
Abrams27: yea, nevertheless in my opinion `"\nevent-end\n"` part can be removed - it made sense with default printing, but with logger it doesn't

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
Abrams27: good enough for me, but I was thinking about an entire if section, as well for similar below

Abrams27: it applies to if sections below too

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: Yeah, I know, there's a lot to extract there, but I left those 3 methods because they're too messy and in many places I know that a method should be extracted but I don't know what this piece of code does precisely and I can't find names for them xd

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: same as above, not sure about the names

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: same as above, not sure about the names

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
Abrams27: cool, tomorrow I will try to look at this part and I will share my thoughts and ideas for naming

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: How exactly?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: > Good job, few things could be changed in my opinion, but in general looks great

Thanks a lot for the detailed review!

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,139 @@
...
+
+    Diagnostic diagnostic = new Diagnostic(range, diagProto.getMessage());
+    if (severity != null) {
+      diagnostic.setSeverity(severity);
+    }
+
+    return diagnostic;
+  }
+
+  private DiagnosticSeverity convertSeverity(Diagnostics.Severity protoSeverity) {
```
Abrams27, line 119: Accidentally  deleted previous comment, but there is my answer:
it is a simple mapping `Diagnostics.Severity` -> `DiagnosticSeverity` what means that it can be represented by `Map<Diagnostics.Severity, DiagnosticSeverity>`

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: You mean wrapping the if in something like _tryProcessFinishedEvent_?

Abrams27: try can suggest `try - catch` block but in general yes:
```
foo(event) {
  if (event.hasStarted() && event.getStarted().getCommand().equals("build")) {
    processedStartedEvent(event.getStarted());
  }
}
``` 
and then replace this code with foo(event)

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: I agree but here I was not sure about the logic (and naming). What exactly does the code inside _if (bspClient != null) {_ do?

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: Right, good idea

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
Abrams27: In my opinion this predicate is always true

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: Should I make all the logs like this: LOGGER.info("Got event {} event-end", event) (I mean the {})?

Abrams27: yes, even with `LOGGER.info("Got event {}", event) `

src/main/java/org/jetbrains/bsp/bazel/server/BepStreamObserver.java
gerardd33: not sure about the logic/name again

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: I think it's unnecessary too, but I didn't want to break it. Still, I mean this:

`
for (List<PublishDiagnosticsParams> values : filesToDiagnostics.values()) {
          for (PublishDiagnosticsParams param : values) {
            bspClient.onBuildPublishDiagnostics(param);
          }
        }
`

What exactly is happening here and why do we need these three nested loops?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: I wanted to change it like this at first too (after seeing the usage in BspServer) but I think it makes sense this way. In other places it's used to go over many sources and gather the diagnostics from each one (see its usage in a loop in processActionEvent).

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: I tried to do it but this code looks suspicious. Why in a loop over _log_ is there something like this (_target_ and _diagnosticsProtosLocations_ are things from outside the loop):

```
if (hasDiagnosticsOutput) {
        diagnosticsProtosLocations.remove(target.getUri());
      }
```

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+    Optional<DiagnosticSeverity> severity =
+        Optional.ofNullable(CONVERTED_SEVERITY.get(diagProto.getSeverity()));
+
+    Position startPosition =
+        new Position(
+            diagProto.getRange().getStart().getLine(),
+            diagProto.getRange().getStart().getCharacter());
+    Position endPosition =
+        new Position(
+            diagProto.getRange().getEnd().getLine(), diagProto.getRange().getEnd().getCharacter());
```
Abrams27, line 121: add new line after `,`

Abrams27 (changes requested):

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +86,146 @@ public void publishLifecycleEvent(
...
-            String message = progress.getStderr().trim();
-            if (!message.isEmpty()) {
-              buildClientLogger.logMessage(progress.getStderr().trim());
-            }
-          }
+  public void handleEvent(BuildEvent buildEvent) {
+    try {
+      processEvent(buildEvent);
+    } catch (IOException e) {
+      LOGGER.error("Error deserializing BEP proto: {}", e.toString());
```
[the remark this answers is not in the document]
gerardd33: IntelliJ didn't like it and marked it yellow with some error message.

src/main/java/org/jetbrains/bsp/bazel/common/Constants.java
```diff
@@ -1,19 +1,29 @@
...
   public static final String BAZELBSP_LOG_FILE_NAME = "bazelbsp.log";
   public static final String BAZELBSP_TRACE_JSON_FILE_NAME = "bazelbsp.trace.json";
+
+  public static final String SCALAC = "Scalac";
+  public static final String KOTLINC = "KotlinCompile";
+  public static final String JAVAC = "Javac";
+
+  public static final List<String> SUPPORTED_LANGUAGES =
+      ImmutableList.of("scala", "java", "kotlin");
+  public static final Set<String> SUPPORTED_ACTIONS = ImmutableSet.of(KOTLINC, JAVAC, SCALAC);
```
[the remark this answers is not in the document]
agluszak: also why is SUPPORTED_LANGUAGES a list and SUPPORTED_ACTIONS a set?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +86,146 @@ public void publishLifecycleEvent(
...
-              return;
-            } else if (taskParkingLot.size() > 1) {
-              System.out.println("More than 1 start event was found");
-              return;
-            }
+  public void collectDiagnostics(
+      Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics,
+      BuildTargetIdentifier target,
+      String diagnosticsLocation)
+      throws IOException {
```
agluszak, line 96: Ideally we'd like to get rid of checked exceptions, so you can wrap this error *inside* `diagnosticsDispatcher`

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
agluszak: I don't like it - what's the point of having two separate methods `handleEvent` and `processEvent`?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +86,146 @@ public void publishLifecycleEvent(
...
+    } catch (IOException e) {
+      LOGGER.error("Error deserializing BEP proto: {}", e.toString());
+    }
+  }
 
-        } catch (IOException e) {
-          System.err.println("Error deserializing BEP proto: " + e);
-        }
-      }
+  void processEvent(BuildEvent buildEvent) throws IOException {
```
agluszak, line 113: Imo this method should be deleted (see the comment above), but default visibility is probably not what we want anyway

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +86,146 @@ public void publishLifecycleEvent(
...
+
+  private void processCompletedEventIfPresent(BuildEventStreamProtos.BuildEvent event)
+      throws IOException {
+    if (event.hasCompleted()) {
+      processCompletedEvent(event.getCompleted());
+    }
+  }
+
+  private void processActionEventIfPresent(BuildEventStreamProtos.BuildEvent event)
+      throws IOException {
```
agluszak, line 154: why does it throw?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +86,146 @@ public void publishLifecycleEvent(
...
+
+    return new PublishDiagnosticsParams(
+        new TextDocumentIdentifier(Uri.fromAbsolutePath(fileLocation).toString()),
+        new BuildTargetIdentifier(""),
+        diagnostics,
+        false);
+  }
+
+  private void processCompletedEvent(BuildEventStreamProtos.TargetComplete targetComplete)
+      throws IOException {
```
agluszak, line 222: Again I think this IO exception should be wrapped inside private methods

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -266,132 +257,106 @@ private void processCompletionEvent(BuildEventStreamProtos.BuildEvent event) thr
...
-    String actionType = action.getType();
-    if (!SUPPORTED_ACTIONS.contains(actionType)) {
+  private void processAbortedEvent(BuildEventStreamProtos.Aborted aborted) {
+    if (aborted.getReason() != BuildEventStreamProtos.Aborted.AbortReason.NO_BUILD) {
+      buildClientLogger.logError(
+          "Command aborted with reason " + aborted.getReason() + ": " + aborted.getDescription());
+    }
+  }
+
+  private void processActionEvent(BuildEventStreamProtos.ActionExecuted action) throws IOException {
```
agluszak, line 267: same as above

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -266,132 +257,106 @@ private void processCompletionEvent(BuildEventStreamProtos.BuildEvent event) thr
...
-                  new Position(
-                      diagProto.getRange().getStart().getLine(),
-                      diagProto.getRange().getStart().getCharacter()),
+                int urlEnd = fileInfo.indexOf(erroredFile) + erroredFile.length();
+                fileLocation = fileInfo.substring(0, urlEnd);
+                lineLocation = fileInfo.substring(urlEnd + 1).split(":");
+              } else {
+                int urlEnd = error.indexOf(erroredFile) + erroredFile.length();
+                fileLocation = error.substring(error.indexOf("ERROR: ") + 7, urlEnd);
+                lineLocation = error.substring(urlEnd + 1).split("(:)|( )");
```
agluszak, line 331: this logic DEFINITELY deserves some unit testing

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -266,132 +257,106 @@ private void processCompletionEvent(BuildEventStreamProtos.BuildEvent event) thr
...
+              diagnostics.add(diagnostic);
+
+              fileDiagnostics.put(fileLocation, diagnostics);
+            });
+
+    return fileDiagnostics;
   }
 
-  public TreeSet<Uri> fetchScalacClasspath() {
+  public TreeSet<Uri> getCompilerClasspath() {
```
agluszak, line 355: return type can be changed to Set I guess

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+  public DiagnosticsDispatcher(BazelBspServer bspServer, BuildClient bspClient) {
+    this.bspServer = bspServer;
+    this.bspClient = bspClient;
+  }
+
+  public void collectDiagnostics(
+      Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics,
+      BuildTargetIdentifier target,
+      String diagnosticsLocation)
+      throws IOException {
```
agluszak, line 53: another expection to be wrapped

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+
+  public void emitDiagnostics(
+      Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics, BuildTargetIdentifier target) {
+    for (SourceItem source : bspServer.getCachedBuildTargetSources(target)) {
+      Uri sourceUri = Uri.fromFileUri(source.getUri());
+      if (!filesToDiagnostics.containsKey(sourceUri)) {
+        addSource(filesToDiagnostics, sourceUri, target);
+      }
+
+      if (bspClient != null) {
```
agluszak, line 73: can bspClient ever be null? does it make sense to check it at this point?

agluszak (commented):
Good job, it requires some improvements though - maybe @andrefmrocha will explain the less clear parts of the logic

src/main/java/org/jetbrains/bsp/bazel/common/Constants.java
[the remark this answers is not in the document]
gerardd33: I just moved what was already in the class to constants, thanks for noticing, I'll change it.

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: Visibility was an accident, sorry. As to having this extracted method, it was needed when this block in the try-catch was larger but now that the ifs are extracted _processEvent_ is much shorter so maybe you're right about not extracting it.

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: Yeah, I also thought it didn't, I'll remove it.

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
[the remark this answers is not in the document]
jastice: Yeah, it's generally preferable to pass the actual string you intend to be shown, rather than rely on implicit toString

gerardd33:
> Good job, it requires some improvements though - maybe @andrefmrocha will explain the less clear parts of the logic

We sat yesterday with @Abrams27 and analysed all of those most messy/complicated methods, so I'll be able to refactor them too after the weekend. But there will indeed still be several questions for André, for example [here](https://github.com/JetBrains/bazel-bsp/pull/21#discussion_r534582423).

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: How would you do it otherwise? The try-catch in handleEvent aggregates and handles all possible exceptions from the first _parseFrom_ and event processors (which are essentially also parsing functions). If I were to wrap everything locally with try-catch in these processors, I'd have to have 5 of them (in _processActionEvent_ specifically _collectDiagnostics_ throw).

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: How exactly?

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: This google formatting does that XD (splits _startPosition_ and takes back _endPosition_ into one line).

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: I'll try to wrap all the IOExceptions while refactoring those methods more thoroughly later.

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
andrefmrocha: > In my opinion this predicate is always true
I'm not sure how that has changed but since the BSP server needed access to the BEP client and the BEP client to the BSP client  (provided by the BSP server), there was at that point a race condition where an event could have been emitted before BEP server had access to it.

> What exactly is happening here and why do we need these three nested loops?
This is a bit of a more messy part of the code since this is to ensure backwards compatibility with older versions of bazel. Basically here, what happens is that diagnostics events coming from the BEP, after being converted to BSP diagnostics, have to be published to the client. It's three nested loops due to this: 
- a target has multiple diagnostics (usually one per source file)
- a  source file might have multiple diagnostics
- a BEP diagnostic event might have more than one BSP diagnostic

This can be flattened, but, other than, not sure if there are any improvements that can be done.

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
andrefmrocha: This is due to the backwards compatibility with previous bazel versions. We try to cache the location of the diagnostic files based on a query we can do to bazel.

However, if a diagnostic log comes over the BEP, it means that that file a newer version of bazel is being used, therefore, we can delete the cache of the diagnostics file we got beforehand. To do this in a more functional way, you can probably look into filtering for diagnostics with the name `diagnostics` and then if there are any, removing the cache based on the `hasDIagnosticsOutput` flag . Wdyt?

gerardd33: Okay, so in the current code

``
if (hasDiagnosticsOutput) {
        diagnosticsProtosLocations.remove(target.getUri());
      }
``

can just be removed from the loop and put at the end of the method?

andrefmrocha: It must only be executed if there's any diagnostics log but yes

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: > I'm not sure how that has changed but since the BSP server needed access to the BEP client and the BEP client to the BSP client (provided by the BSP server), there was at that point a race condition where an event could have been emitted before BEP server had access to it.

There is indeed something like this in _Server.java_ but there we have:

``
BepServer bepServer = new BepServer(bspServer, launcher.getRemoteProxy(), buildClientLogger);
bspServer.setBepServer(bepServer);
`` 
(where ``launcher.getRemoteProxy()`` is ``BuildClient bspClient``)

so I think it's not a problem, maybe only if this _launcher.getRemoteProxy()_ works in some strange way. What do you think?

> This is a bit of a more messy part of the code since this is to ensure backwards compatibility with older versions of bazel. Basically here, what happens is that diagnostics events coming from the BEP, after being converted to BSP diagnostics, have to be published to the client. It's three nested loops due to this:
> 
> * a target has multiple diagnostics (usually one per source file)
> * a  source file might have multiple diagnostics
> * a BEP diagnostic event might have more than one BSP diagnostic
> 
> This can be flattened, but, other than, not sure if there are any improvements that can be done.

Thanks for explaining! But still what we do here is we put things into _filesToDiagnostics_, then do these two loops over _filesToDiagnostics.values()_ to publish and then go to the next source without removing anything from _filesToDiagnostics_. So aren't some parts of _filesToDiagnostics_ published multiple times? Can't these two loops be moved out and after the one over the sources or am I misunderstanding this?

src/main/java/org/jetbrains/bsp/bazel/server/BepStreamObserver.java
```diff
@@ -0,0 +1,55 @@
...
+    responseObserver.onCompleted();
+  }
+
+  private boolean isRequestBazelBuildEvent(PublishBuildToolEventStreamRequest request) {
+    return request
+        .getOrderedBuildEvent()
+        .getEvent()
+        .getBazelEvent()
+        .getTypeUrl()
+        .equals("type.googleapis.com/build_event_stream.BuildEvent");
```
Abrams27, line 53: I think that this literal could be moved to the separate variable

Abrams27 (changes requested):

src/main/java/org/jetbrains/bsp/bazel/server/ScalaBspServer.java
```diff
@@ -34,21 +35,13 @@
   private final TargetsResolver targetsResolver;
   private final ActionGraphResolver actionGraphResolver;
 
-  private final String scalac;
-  private final String javac;
   private final String execRoot;
 
   // TODO: too many arguments!!, constants will be moved
```
agluszak, line 40: TODO no longer valid :)

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
andrefmrocha: I believe that the `filesToDiagnostics` has their values removed once another event for that target is received. We can't just remove them because a specific target might be cached and we still need to resend all diagnostics when a new compilation happens.
This leads to the same diagnostic having the possibility of being sent more than once during a single compilation yes (which is inefficient but is assumed as a possibility in the protocol), but your change would make us not be protocol-compliant. I believe there are ways of making this more efficient but they are more complicated and probably should not be addressed in this PR. Wdyt?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -178,6 +143,12 @@ private void processedStartedEvent(BuildEventStreamProtos.BuildStarted buildStar
     taskParkingLot.add(taskId);
   }
 
+  private void processFinishedEventIfPresent(BuildEventStreamProtos.BuildEvent event) {
+    if (event.hasFinished()) {
+      processFinishedEvent(event.getFinished());
```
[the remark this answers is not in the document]
gerardd33: You mean just moving the body of ``processFinishedEvent`` into ``if (event.hasFinished()) {``?

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: Yup, I agree. Just wanted to understand what was going on there because it didn't make sense at first sight.

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: Wait, then why is it removed anyway after collecting the diagnostics inside ``if (filesToDiagnostics.isEmpty() && hasDiagnosticsOutput)``?

andrefmrocha: If a diagnostics event for a given target has been found but it is empty, then it means that the all of the cached information is now obsolete and that we have no need to go over it with the information from the `diagnosticsProtosLocation`

gerardd33: What I mean is, in the original code:

1. If there are some diagnostic logs, then we remove it in the loop.
2. If there aren't any, then filesToDiagnostics is empty, so we remove it anyway at the end.

So I think it can be removed anyway at the end iff ``hasDiagnosticsOutput`` is true.

I've just pushed my version with this change. Could you take a look?

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -40,30 +35,34 @@
...
-  private final BazelBspServer bspServer;
+  private static final Logger LOGGER = LogManager.getLogger(BepServer.class);
+
   private final BuildClient bspClient;
+  private final BuildClientLogger buildClientLogger;
+
+  private final BazelBspServer bspServer;
+  private final DiagnosticsDispatcher diagnosticsDispatcher;
+
+  private final Stack<TaskId> taskParkingLot = new Stack<>();
```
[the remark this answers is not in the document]
gerardd33: I renamed it and changed it to Deque because Stack is legacy, but what exactly is your idea with the variables? The only way I see it to have two attributes: startedEventTaskId and _uniqueStartedEvent_ and then having something very ugly like:

```
if (startedEventTaskId == null) {
startedEventTaskId = taskId;
} else {
_uniqueStartedEvent_ = true;
}
```

(or the same thing with some Optional but it doesn't look better imo)

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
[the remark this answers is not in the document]
agluszak: yes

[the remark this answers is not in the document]
Abrams27: I disagree - this logic is too complex for being inside the `if` statement - it should be separated to the method. Nevertheless I would consider better names - `processFinishedEvent` and `processFinishedEventForFinishedEvent` or something similar to show that this method should be called only for the finished event.

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
andrefmrocha: Looks good! Good job!

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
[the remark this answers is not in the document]
andrefmrocha: Maybe I'd suggest using `processEvent` for processing if they're present or not and `consumeEvent` for the actual logic of the event. Wdyt? I think consuming already underlies that it must be present.

src/main/java/org/jetbrains/bsp/bazel/server/BepStreamObserver.java
```diff
@@ -0,0 +1,57 @@
...
+import com.google.devtools.build.v1.PublishBuildToolEventStreamRequest;
+import com.google.devtools.build.v1.PublishBuildToolEventStreamResponse;
+import io.grpc.stub.StreamObserver;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class BepStreamObserver implements StreamObserver<PublishBuildToolEventStreamRequest> {
+
+  private static final Logger LOGGER = LogManager.getLogger(BepStreamObserver.class);
+  private static final String BUILD_EVENT_TYPE_URL =
```
andrefmrocha, line 12: This can be moved to the constants file

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -193,68 +189,69 @@ private void processCompletedEvent(BuildEventStreamProtos.TargetComplete targetC
...
-        diagnosticsProtosLocations.remove(target.getUri());
+    Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics =
+        action.getActionMetadataLogsList().stream()
+            .filter(log -> log.getName().equals("diagnostics"))
+            .peek(log -> LOGGER.info("Found diagnostics file in {}", log.getUri()))
+            .map(log -> diagnosticsDispatcher.collectDiagnostics(target, log.getUri().substring(7)))
+            .collect(HashMap::new, Map::putAll, Map::putAll);
+
+    if (hasDiagnosticsOutput(target)) {
+      if (filesToDiagnostics.isEmpty()) {
```
agluszak, line 240: Why did you separate that if with && into two ifs? This changed the semantics https://github.com/JetBrains/bazel-bsp/blob/dev/src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java#L292

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -610,14 +608,9 @@ private String getSourcesRoot(String uri) {
     for (Map.Entry<String, String> diagnostics : diagnosticsProtosLocations.entrySet()) {
```
agluszak, line 608: it can be streamified

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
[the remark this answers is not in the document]
gerardd33: I also think unsplitting it is a bad idea, it's much harder to read code with such nested ifs and easier if you have a properly named method. As to the names, I used ``*ifPresent`` because for each type of event we check if ``event`` _has_ it. I think it's more obvious from outside what happens here. But André's idea is also fine, so I can change it to this if you want. It would be like this, right?:

```
 private void processActionEvent(BuildEventStreamProtos.BuildEvent event) {
    if (event.hasAction()) {
      consumeActionEvent(event.getAction());
    }
  }
```

andrefmrocha (changes requested):
Just some minor additions, I think the vital questions have already been asked :smiley:

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,124 @@
...
+                  diagnostics -> getUriForPath(diagnostics.getPath()),
+                  diagnostics -> convertDiagnostics(target, diagnostics)));
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public void emitDiagnostics(
+      Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics, BuildTargetIdentifier target) {
+    for (SourceItem source : bspServer.getCachedBuildTargetSources(target)) {
```
andrefmrocha, line 70: Can be streamified

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -88,156 +88,96 @@ public void publishLifecycleEvent(
...
-    List<BuildEventStreamProtos.OutputGroup> outputGroups =
-        event.getCompleted().getOutputGroupList();
+  private void processCompletedEvent(BuildEventStreamProtos.TargetComplete targetComplete) {
+    List<OutputGroup> outputGroups = targetComplete.getOutputGroupList();
+
     if (outputGroups.size() == 1) {
       BuildEventStreamProtos.OutputGroup outputGroup = outputGroups.get(0);
-      if ("scala_compiler_classpath_files".equals(outputGroup.getName())) {
+
+      if (outputGroup.getName().equals("scala_compiler_classpath_files")) {
```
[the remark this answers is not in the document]
gerardd33: Right, _parseStdErrDiagnostics_ and _processCompletedEvent_ are the only ones not refactored yet, I'll extract some constants there.

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -51,8 +52,8 @@
   private final BazelBspServer bspServer;
   private final DiagnosticsDispatcher diagnosticsDispatcher;
 
-  private final Stack<TaskId> taskParkingLot = new Stack<>();
   private final Set<Uri> compilerClasspath = new TreeSet<>();
+  private final Deque<TaskId> startedEventTaskIds = new ArrayDeque<>();
```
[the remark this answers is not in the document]
gerardd33: https://stackoverflow.com/questions/12524826/why-should-i-use-deque-over-stack

It can be any collection, actually it would be best not to have any collection here if it is allowed only to have one element but [I don't know how to do it](https://github.com/JetBrains/bazel-bsp/pull/21#discussion_r537079701).

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
gerardd33: It's not the only change there. I talked [about it](https://github.com/JetBrains/bazel-bsp/pull/21#discussion_r534587407) with André.

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
gerardd33: Right, but it's _BazelBspServer_, there's much more to be done there later.

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: Done, but I'm not sure if it can be done without this big forEach function, because we must publish everything from filesToDiagnostics each time directly after adding each source, right?

andrefmrocha: I don't think it can be better

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
Abrams27: an not in this PR

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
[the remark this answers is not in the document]
Abrams27: I think that this literal is specific for this class not for entire project - it should be moved to the static final field but in this class.

src/main/java/org/jetbrains/bsp/bazel/server/ScalaBspServer.java
```diff
@@ -63,7 +55,8 @@ public ScalaBspServer(
         targetsResolver.getTargetsOptions(targetsUnion, "scalacopts");
     ActionGraphParser actionGraphParser =
         actionGraphResolver.parseActionGraph(
-            MnemonicsUtils.getMnemonics(targetsUnion, Lists.newArrayList(scalac, javac)));
+            MnemonicsUtils.getMnemonics(
+                targetsUnion, Lists.newArrayList(Constants.SCALAC, Constants.JAVAC)));
```
Abrams27, line 59: is this array modified? if not then it would be better to use `ImmutableList.of(...)`

src/main/java/org/jetbrains/bsp/bazel/server/JavaBspServer.java
```diff
@@ -59,7 +50,8 @@ public JavaBspServer(
     // TODO(andrefmrocha): Remove this when kotlin is natively supported
     ActionGraphParser actionGraphParser =
         actionGraphResolver.parseActionGraph(
-            MnemonicsUtils.getMnemonics(targetsUnion, Lists.newArrayList(javac, kotlinc)));
+            MnemonicsUtils.getMnemonics(
```
Abrams27, line 53: is this array modified? if not then it would be better to use ImmutableList.of(...)

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.stream.Collectors;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jetbrains.bsp.bazel.common.Uri;
+
+public class DiagnosticsDispatcher {
+
+  private static final Logger LOGGER = LogManager.getLogger(DiagnosticsDispatcher.class);
+
+  private final Map<Severity, DiagnosticSeverity> CONVERTED_SEVERITY =
```
Abrams27, line 34: static

src/main/java/org/jetbrains/bsp/bazel/server/BepStreamObserver.java
Abrams27: It's specific for this class, not entire project - in my opinion it should stay in this class

Abrams27 (commented):
lets go lets go

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
gerardd33: Yeah, that's what I meant

gerardd33:
Okay, I've finally finished all of the major tasks in this PR. These last processing methods were very challenging so there will surely be several things to fix/improve, but in general there shouldn't be more large changes. (Btw later I'll try to create some packages to tidy up)

Two more questions for @andrefmrocha:

- I marked several places where I wasn't sure about the logic in particular as "TODO FOR REVIEW", could you take a look at them?
- What about this _convertExitCode_? — What does _convertExitCode_ do exactly, why is it in BepServer (it's only used from BspServer) and where to move it if needed?

andrefmrocha:
> I marked several places where I wasn't sure about the logic in particular as "TODO FOR REVIEW", could you take a look at them?

Will take a look at it once again today.

> What about this convertExitCode? — What does convertExitCode do exactly, why is it in BepServer (it's only used from BspServer) and where to move it if needed?

The location can probably be refactored as well (it was used at further places and then deleted somewhere along the way) but this is related to converting Bazel compilation exit codes to BSP compliant ones. This is BazelToBsp conversion logic though

src/main/java/org/jetbrains/bsp/bazel/server/BepServer.java
```diff
@@ -165,62 +158,43 @@ private void processFinishedEvent(BuildEventStreamProtos.BuildFinished buildFini
...
-                }
+      OutputGroup outputGroup = outputGroups.get(0);
+      if (outputGroup.getName().equals(Constants.SCALA_COMPILER_CLASSPATH_FILES)) {
+        processFileSets(outputGroup);
+      }
+    }
+  }
 
+  // TODO FOR REVIEW: better naming suggestions?
+  private void processFileSets(OutputGroup outputGroup) {
```
andrefmrocha, line 178: This is used to fetch the scala compiler classpath. Perhaps calling it `fetchScalaJars` , `getScalaJars`, `getScalaClasspath` sound more interesting.

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
```diff
@@ -67,18 +67,24 @@ public DiagnosticsDispatcher(BazelBspServer bspServer, BuildClient bspClient) {
 
   public void emitDiagnostics(
       Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics, BuildTargetIdentifier target) {
-    for (SourceItem source : bspServer.getCachedBuildTargetSources(target)) {
-      Uri sourceUri = Uri.fromFileUri(source.getUri());
-      PublishDiagnosticsParams publishDiagnosticsParams =
-          new PublishDiagnosticsParams(
-              new TextDocumentIdentifier(sourceUri.toString()), target, new ArrayList<>(), true);
+    Streams.stream(bspServer.getCachedBuildTargetSources(target))
```
andrefmrocha, line 70: Any reason to not call `bspServer.getCachedBuildTargetSources(target).stream()` instead of `Streams.stream()`?

src/main/java/org/jetbrains/bsp/bazel/server/ParsingUtils.java
```diff
@@ -0,0 +1,125 @@
...
+  public static URI parseUri(String uri) {
+    try {
+      return new URI(uri);
+    } catch (URISyntaxException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  // TODO FOR REVIEW: naming suggestions? what exactly is the thing that it returns?
+  public static List<String> parseClasspathFromFile(URI protoPathUri) {
```
andrefmrocha, line 40: This is obtaining directly the jars location from a bazel aspect. I would suggest calling the method `parseClasspathFromAspect` and the variable `dependenciesAspectOutput` (URI is already outlined by the type)

src/main/java/org/jetbrains/bsp/bazel/server/ParsingUtils.java
```diff
@@ -0,0 +1,125 @@
...
+    try {
+      return Files.readLines(new File(protoPathUri), StandardCharsets.UTF_8).stream()
+          .map(line -> Splitter.on("\"").splitToList(line))
+          .peek(
+              parts -> {
+                if (parts.size() != 3) {
+                  // TODO FOR REVIEW: verify the slight change of semantics
+                  // TODO is it OK to throw an exception at the beginning if any line is corrupted?
```
andrefmrocha, line 48: If this by any chance is not fine we probably should handle this better actually. I never actually played around with this logic, but I'm not sure its a good idea for us to just break everything upon this breaking.
The refactor is indeed correct but this should probably just filter out those broken dependencies and signal the error.
Probably this out of the scope of this PR though.

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -567,7 +571,7 @@ private String getSourcesRoot(String uri) {
 
   private Either<ResponseError, CompileResult> buildTargetsWithBep(
       List<BuildTargetIdentifier> targets, List<String> extraFlags) {
-    List<String> args = Lists.newArrayList("build");
+    List<String> args = Lists.newArrayList(Constants.BAZEL_BUILD_COMMAND);
```
andrefmrocha, line 574: Perhaps substitute with `ImmutableList.of()`?

andrefmrocha (changes requested):
Some minor suggestions :thinking:

src/main/java/org/jetbrains/bsp/bazel/server/DiagnosticsDispatcher.java
gerardd33: I did it because the method returned Iterable and Iterable does not have ``.stream()``, but I didn't notice it's just a method from BazelBspServer and not something external so I'll just change the return type from Iterable to Collection there.

src/main/java/org/jetbrains/bsp/bazel/server/ParsingUtils.java
andrefmrocha: Perhaps we could then to create an issue out of this thread?

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
gerardd33: I agree, but there are another 15 random newArrayList and new ArrayList<>() in BazelBspServer so probably for another PR.

andrefmrocha: Sure thing.

src/main/java/org/jetbrains/bsp/bazel/server/ParsingUtils.java
gerardd33: Good idea

gerardd33:
> Will take a look at it once again today.

Thanks for the suggestions!

> The location can probably be refactored as well (it was used at further places and then deleted somewhere along the way) but this is related to converting Bazel compilation exit codes to BSP compliant ones. This is BazelToBsp conversion logic though

I'll move it to BazelBspServer for now. Some things will still be extracted from it later and I think there are some methods with which it will fit well somewhere.

agluszak (approved):
LGTM

Abrams27 (approved):
remember to squash

src/main/java/org/jetbrains/bsp/bazel/server/utils/ParsingUtils.java
```diff
@@ -36,16 +36,24 @@ public static URI parseUri(String uri) {
     }
   }
 
-  // TODO FOR REVIEW: naming suggestions? what exactly is the thing that it returns?
-  public static List<String> parseClasspathFromFile(URI protoPathUri) {
+  public static StatusCode parseExitCode(int exitCode) {
```
andrefmrocha, line 39: Not sure this is the best place for this to leave (I'm always that fond of utils folders tbh) but it's something to think of another time

andrefmrocha (approved):
Let's :rocket: it!

Abrams27:
> remember to squash

: (

andrefmrocha:
> > remember to squash
> 
> : (

Whoops, that's on me, my bad. Why don't you guys enforce it through settings though?

jastice:
They might not have access to settings. what settings do we want exactly?

agluszak:
@jastice yup, I was just going to write that we don't have access to settings. We'd like to have only squash merging enabled (https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuring-commit-squashing-for-pull-requests) and the default brached changed to dev (https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/changing-the-default-branch)

jastice:
done

PR #22, opened 2020-12-16
target scala test classes

Abrams27:
Implementation of `buildTargetScalaTestClasses` method.

**Currently it is a PoC - mainly copied logic from another method - will be changed.**

This logic appears to be used in multiple places (including the further `buildTargetScalaMainClasses` implementation), so I decided to extract it - some refactor is predicted.

**Only logic is needed to be checked - code will be refactored**

### DIFF
--- a/src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServerConfig.java
+++ b/src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServerConfig.java
@@ -23,7 +23,7 @@ public class BazelBspServerConfig {
     List<String> targetProjectPaths =
         args.length == 2
             ? Arrays.asList(args[1].split(","))
-            : Collections.unmodifiableList(Collections.singletonList("//..."));
+            : Collections.singletonList("//...");
 
     return new BazelBspServerConfig(pathToBazel, targetProjectPaths);
   }
--- a/src/main/java/org/jetbrains/bsp/bazel/server/ScalaBspServer.java
+++ b/src/main/java/org/jetbrains/bsp/bazel/server/ScalaBspServer.java
@@ -3,12 +3,14 @@ package org.jetbrains.bsp.bazel.server;
 import ch.epfl.scala.bsp4j.BuildTargetIdentifier;
 import ch.epfl.scala.bsp4j.ScalaMainClassesParams;
 import ch.epfl.scala.bsp4j.ScalaMainClassesResult;
+import ch.epfl.scala.bsp4j.ScalaTestClassesItem;
 import ch.epfl.scala.bsp4j.ScalaTestClassesParams;
 import ch.epfl.scala.bsp4j.ScalaTestClassesResult;
 import ch.epfl.scala.bsp4j.ScalacOptionsItem;
 import ch.epfl.scala.bsp4j.ScalacOptionsParams;
 import ch.epfl.scala.bsp4j.ScalacOptionsResult;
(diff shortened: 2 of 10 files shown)


src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServerConfig.java
```diff
@@ -23,7 +23,7 @@ public static BazelBspServerConfig from(String[] args) {
     List<String> targetProjectPaths =
         args.length == 2
             ? Arrays.asList(args[1].split(","))
-            : Collections.unmodifiableList(Collections.singletonList("//..."));
+            : Collections.singletonList("//...");
```
andrefmrocha, line 26: Any reason for this change to be happening here? Not sure if its under the scope of the PR or if it should be happening as a whole

src/main/java/org/jetbrains/bsp/bazel/server/ScalaBspServer.java
```diff
@@ -21,84 +21,132 @@
...
+                                        && attribute.getExplicitlySpecified()))
+                .map(
+                    rule ->
+                        new ScalaTestClassesItem(
+                            new BuildTargetIdentifier(rule.getName()),
+                            getTestClasses(rule, query)))
+                .collect(Collectors.toList()));
+    return Either.forRight(scalaTestClassesResult);
+  }
+
```
andrefmrocha, line 132: It is probably a good idea to parameterize this query right from the beginning to search for those specific build targets since this can be a very much costly move right off the bat. That would remove the filter that is being done here that is also very costly itself since it tries to match in any given case

src/main/java/org/jetbrains/bsp/bazel/server/ScalaBspServer.java
```diff
@@ -21,84 +21,132 @@
...
+  private List<String> getTestClasses(Build.Rule rule, Build.QueryResult queryResult) {
+    return rule.getAttributeList().stream()
+        .filter(
+            attribute ->
+                attribute.getName().equals("main_class")
+                    && attribute.hasExplicitlySpecified()
+                    && attribute.getExplicitlySpecified())
```
andrefmrocha, line 139: This condition is also used above. Perhaps it can be extracted?

andrefmrocha (changes requested):
Overall looks great, just some minor changes :+1:

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServerConfig.java
Abrams27: yea, I isn't connected with this PR - intellij marked it and I decided to change it - I can revert it

src/main/java/org/jetbrains/bsp/bazel/server/ScalaBspServer.java
Abrams27: great, I will update it here. Also, there are other places where such parameterization should be implemented - I will create a new ticket for it

src/main/java/org/jetbrains/bsp/bazel/server/ScalaBspServer.java
Abrams27: of course

andrefmrocha:
Btw don't forget to fix formatting issues :smile:

PR #23, opened 2020-12-24
bazel runner refactor

Abrams27:
Introduces `BazelRunnerBuilder` - now `bazel` command has to be created with builder.

**resolvers have to be changed in the future!!**

EDIT:
I the future builder should be improved - specific builder methods for query, build, run etc + additional builders for parameters


src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelDataResolver.java
```diff
@@ -1,30 +1,32 @@
...
     Path workspacePath = Paths.get(execRoot);
     String workspaceLabel = workspacePath.toFile().getName();
     return new BazelData(execRoot, workspaceRoot, binRoot, workspaceLabel);
   }
 
-  private String readOnlyBazelLine(String... args) {
-    ProcessResults processResults = bazelRunner.runBazelCommand(args);
+  private String readOnlyBazelLine(String argument) {
+    ProcessResults processResults =
+        bazelRunner.commandBuilder().info().withArgument(argument).runBazel();
```
andrefmrocha, line 29: The naming of this function does not seem intuitive. Perhaps `readBazelInfo` is better, wdyt?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
```diff
@@ -0,0 +1,93 @@
...
+public class BazelRunner {
+
+  private static final Logger LOGGER = LogManager.getLogger(BazelRunner.class);
+
+  private static final String PUBLISH_ALL_ACTIONS = "--build_event_publish_all_actions";
+  private static final String BES_BACKEND = "--bes_backend=grpc://localhost:";
+
+  private final Semaphore processLock;
+  private final String bazel;
+  private Integer besBackendPort;
```
andrefmrocha, line 20: Perhaps we could use an [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) to make sure the access to this variable is safe. Wdyt?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/utils/BazelArgumentsUtils.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private static String getMnemonicsForTargets(String targets, List<String> languageIds) {
+    return languageIds.stream()
+        .filter(Objects::nonNull)
+        .map(languageId -> getMnemonicForLanguageAndTargets(languageId, targets))
+        .collect(Collectors.joining(MNEMONIC_DELIMITER));
+  }
+
+  private static String getMnemonicForLanguageAndTargets(String languageId, String targets) {
+    return String.format("mnemonic(%s, %s)", languageId, targets);
```
andrefmrocha, line 36: I'm not sure about this myself but do you think this string should be extracted?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/utils/BazelArgumentsUtils.java
```diff
@@ -0,0 +1,38 @@
...
+    return getMnemonicsForTargets(joinedTargets, languageIds);
+  }
+
+  private static String joinBazelTargets(List<String> targets) {
+    return Joiner.on(JOIN_TARGETS_DELIMITER).join(targets);
+  }
+
+  private static String getMnemonicsForTargets(String targets, List<String> languageIds) {
+    return languageIds.stream()
+        .filter(Objects::nonNull)
```
andrefmrocha, line 30: Is there any reason for this filter? Are we expecting null results in the list?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunnerBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+    this.arguments = new ArrayList<>();
+  }
+
+  public BazelRunnerBuilder withFlag(BazelRunnerFlag flag) {
+    flags.add(flag.toString());
+
+    return this;
+  }
+
+  public BazelRunnerBuilder withFlag(BazelRunnerFlag flag, String value) {
```
andrefmrocha, line 29: Perhaps we could create methods for very common flags, especially the proto output one.

andrefmrocha (changes requested):
Overall looks great! Just some minor ideas to make it even better :smile:

agluszak (approved):
That builder is 🔥

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -187,16 +184,248 @@ public void onBuildExit() {
...
+                  .withTargets(targets)
+                  .runBazel();
+          Build.QueryResult queryResult = QueryResolver.getQueryResultForProcess(processResults);
+
+          List<SourcesItem> sources =
+              queryResult.getTargetList().stream()
+                  .map(Build.Target::getRule)
+                  .map(
+                      rule -> {
+                        BuildTargetIdentifier label = new BuildTargetIdentifier(rule.getName());
```
agluszak, line 210: this can be extracted to a method I guess?

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -187,16 +184,248 @@ public void onBuildExit() {
...
+                  .map(BuildTargetIdentifier::getUri)
+                  .collect(Collectors.toList());
+
+          DependencySourcesResult result =
+              new DependencySourcesResult(
+                  targets.stream()
+                      .sorted()
+                      .map(
+                          target -> {
+                            List<String> files =
```
agluszak, line 275: same

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -187,16 +184,248 @@ public void onBuildExit() {
...
+              new ResourcesResult(
+                  query.getTargetList().stream()
+                      .map(Build.Target::getRule)
+                      .filter(
+                          rule ->
+                              resourcesParams.getTargets().stream()
+                                  .anyMatch(target -> target.getUri().equals(rule.getName())))
+                      .filter(
+                          rule ->
+                              rule.getAttributeList().stream()
```
agluszak, line 314: same

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/ProcessResults.java
```diff
@@ -26,8 +29,8 @@ public InputStream getStderrStream() {
     return stderr;
   }
 
-  public int getExitCode() {
-    return exitCode;
+  public StatusCode getExitCode() {
```
agluszak, line 32: now the naming is inconsistent: getExitCode returns a StatusCode object.

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelDataResolver.java
Abrams27: In general I had a problem with names there - I wanted to to do a mapping 1-1 from `bazel <command>` to `bazelRunner.commandBuilder().<command>`. Nevertheless I fully agree that it might be not enough intuitive. `readBazelInfo` is difficult to extend into the pattern (for `run` or `build`), maybe `infoCommand`? but then `runCommand` is a bit ambiguous? Any other ideas?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/utils/BazelArgumentsUtils.java
Abrams27: yea, maybe I should even add `.withMnemonic` to the builder?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/utils/BazelArgumentsUtils.java
Abrams27: I copied it, but I think that it is unnecessary - will remove

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunnerBuilder.java
Abrams27: I think that it is unnecessary - I would like to keep builder as simple as possible. Nevertheless `.withProtoOutputFlag` call is almost the same sa `.withFlag(PROTO_OUTPUT_FLAG)`

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
Abrams27: I believe that isn't strictly related to this PR - there are multiple places where changes are required after this PR. I will do this in the separate PR.

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/ProcessResults.java
Abrams27: right

src/main/java/org/jetbrains/bsp/bazel/server/bazel/utils/BazelArgumentsUtils.java
andrefmrocha: Sounds great!

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelDataResolver.java
andrefmrocha: Maybe `readBazelInfoLine`? Not sure if it doesn't suffer from the same problem

Abrams27: I would like to use the same schema to all bazel commands - `run`, `build`, `info` ...

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunnerBuilder.java
andrefmrocha: I think it would provide a simpler api from who will handle it, since you don't have to worry about the flags and where the imports are to them are. However, I can see that this might bring out some ambiguity to the calls. Maybe a third opinion would be interesting, like @agluszak 's

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelDataResolver.java
andrefmrocha: maybe something like `infoFetch` then? sounds odd but along these lines maybe? definitely a more troublesome name.

Abrams27: Maybe I wasn't clear - I would like to find such schema: `<schema><bazel command><schema>` for example `callBuild` / `callRun`, I think that `runFetch` doesn't sound good

andrefmrocha: Ah, I got it now, sorry. Not sure I can think of a name for this one :thinking: ...

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunnerBuilder.java
agluszak: My initial thought was to have the builder for custom calls and create methods for the most common uses, but after a closer look there aren't any places where the builder is used with the same params more then once :D I agree with Marcin that the builder itself should be kept simple

src/main/java/org/jetbrains/bsp/bazel/server/service/BuildServerService.java
```diff
@@ -220,7 +231,16 @@ public void onBuildExit() {
 
   public Either<ResponseError, ResourcesResult> buildTargetResources(
       ResourcesParams resourcesParams) {
-    Build.QueryResult query = queryResolver.getQuery("query", "--output=proto", "//...");
+    BazelProcessResult bazelProcessResult =
+        bazelRunner
+            .commandBuilder()
+            .query()
+            .withFlag(BazelRunnerFlag.OUTPUT_PROTO)
+            .withArgument("//...")
```
gerardd33, line 239: Maybe we should extract it as a constant?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
```diff
@@ -181,22 +191,22 @@ private JvmBuildTarget getJVMBuildTarget() {
...
-                + ")");
+    BazelProcessResult bazelProcessResult =
+        bazelRunner
+            .commandBuilder()
+            .query()
+            .withFlag(BazelRunnerFlag.OUTPUT_PROTO)
+            .withTargets(bazelTargets)
+            .executeBazelCommand();
+
+    Build.QueryResult queryResult = QueryResolver.getQueryResultForProcess(bazelProcessResult);
```
gerardd33, line 209: Why is ``QueryResolver`` still around only with this one method that is just a wrapped _try-catch_ over ``Build.QueryResult::parseFrom``? I would remove it.

gerardd33 (changes requested):
Great PR, really love the builder!
I have only a few improvement suggestions.

src/main/java/org/jetbrains/bsp/bazel/server/service/BuildServerService.java
Abrams27: this call should be changed, but not in this PR

src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
Abrams27: it needs to be changed, but not in this PR, in general flow should be updated (including resolvers)

src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
```diff
@@ -181,22 +191,22 @@ private JvmBuildTarget getJVMBuildTarget() {
...
   public Either<ResponseError, CompileResult> buildTargetsWithBep(
       List<BuildTargetIdentifier> targets, List<String> extraFlags) {
-    List<String> args = Lists.newArrayList(Constants.BAZEL_BUILD_COMMAND);
-    args.addAll(targets.stream().map(BuildTargetIdentifier::getUri).collect(Collectors.toList()));
-    args.addAll(extraFlags);
-    int exitCode = -1;
+    List<String> bazelTargets =
+        targets.stream().map(BuildTargetIdentifier::getUri).collect(Collectors.toList());
+
+    StatusCode exitCode = StatusCode.ERROR;
```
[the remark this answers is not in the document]
Abrams27: it is not my logic, this variable is updated in try - needs change, but not in this PR

src/main/java/org/jetbrains/bsp/bazel/server/BUILD
```diff
@@ -6,9 +6,11 @@ java_library(
     visibility = ["//visibility:public"],
     deps = [
         "//src/main/java/org/jetbrains/bsp/bazel/common",
+        "//src/main/java/org/jetbrains/bsp/bazel/server/bazel",
+        "//src/main/java/org/jetbrains/bsp/bazel/server/bazel/data:bazel-data",
+        "//src/main/java/org/jetbrains/bsp/bazel/server/bazel/utils:bazel-utils",
```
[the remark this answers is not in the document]
Abrams27: tbh I used to use a plural form for packages, I think that we have to agree on one convention + @agluszak what is the preferred bazel package naming?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelDataResolver.java
```diff
@@ -0,0 +1,33 @@
...
+  private final BazelRunner bazelRunner;
+
+  public BazelDataResolver(BazelRunner bazelRunner) {
+    this.bazelRunner = bazelRunner;
+  }
+
+  public BazelData resolveBazelData() {
+    String execRoot = readOnlyBazelLine("execution_root");
+    String workspaceRoot = readOnlyBazelLine("workspace");
+    String binRoot = readOnlyBazelLine("bazel-bin");
```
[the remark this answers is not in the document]
Abrams27: right

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunnerBuilder.java
andrefmrocha: Alright let's leave it as is :smile:

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunnerBuilder.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  public BazelRunnerBuilder withArgument(String argument) {
+    arguments.add(argument);
+
+    return this;
+  }
+
+  public BazelRunnerBuilder withArguments(List<String> arguments) {
+    this.arguments.addAll(arguments);
```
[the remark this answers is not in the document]
andrefmrocha: This can't exactly be done due to the argument's name. Unless you also suggest to change the argument's name

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+
+  private BazelProcessResult runBazelProcess(
+      String command, List<String> flags, List<String> arguments)
+      throws InterruptedException, IOException {
+    List<String> processArgs = getProcessArgs(command, flags, arguments);
+    LOGGER.info("Running: {}", processArgs);
+
+    processLock.acquire();
```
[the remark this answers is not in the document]
Abrams27: right, I believe that I is a legacy of pre-refactor times

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private List<String> getProcessArgs(String command, List<String> flags, List<String> arguments) {
+    List<String> processArgs = new ArrayList<>();
+
+    processArgs.add(bazel);
+    processArgs.add(command);
+    processArgs.addAll(flags);
+    processArgs.addAll(arguments);
+
+    return processArgs;
```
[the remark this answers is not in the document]
Abrams27: proposition?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunnerBuilder.java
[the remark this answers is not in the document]
Abrams27: will update name

andrefmrocha (approved):
Looks great! Gj!

src/main/java/org/jetbrains/bsp/bazel/server/BUILD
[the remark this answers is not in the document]
gerardd33: I like ``services`` much less than ``service`` but I guess both are fine. We can decide which one we want and adjust the other packages (because now we have one ``util`` and one ``utils``).

src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
gerardd33: Okay, I thought ``QueryResolver`` was larger and your PR disassembled it more.

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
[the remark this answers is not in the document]
gerardd33: Ok, my original idea turned out to be an overkill, sorry xd But you can still do:

```
List<String> processArgs = Lists.newArrayList(bazel, command);
processArgs.addAll(flags);
processArgs.addAll(arguments);
```

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
```diff
@@ -0,0 +1,92 @@
...
+
+  BazelProcessResult runBazelCommandBes(
+      String command, List<String> flags, List<String> arguments) {
+
+    List<String> newFlags = getBesFlags(flags);
+    return runBazelCommand(command, newFlags, arguments);
+  }
+
+  private List<String> getBesFlags(List<String> flags) {
+    List<String> newFlags = new ArrayList<>();
```
gerardd33, line 38: Actually here too you can use ``Lists.newArrayList``.

src/main/java/org/jetbrains/bsp/bazel/server/BUILD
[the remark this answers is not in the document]
agluszak: I agree with 1, but prefer plurals for packages

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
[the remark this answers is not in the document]
agluszak: @Abrams27 you introduced that semaphore yourself as far as I remember xD it was to replace something else, but not a synchronized method. I'm not sure which one I prefer tbh

src/main/java/org/jetbrains/bsp/bazel/server/BUILD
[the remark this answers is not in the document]
agluszak: in general we want to use implicitly generated targets for packages

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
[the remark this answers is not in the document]
Abrams27: right, forgot xd

src/main/java/org/jetbrains/bsp/bazel/server/BUILD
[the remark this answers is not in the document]
Abrams27: done (used plural)

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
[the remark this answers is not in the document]
Abrams27: changed to `synchronized`

src/main/java/org/jetbrains/bsp/bazel/server/bazel/BazelRunner.java
Abrams27: done

gerardd33 (approved):
All looks good

PR #30, opened 2021-01-03
Packages rearrange + Utils refactor

### LINKED ISSUE #27: Moving methods from ParsingUtils closer to where they're used
Suggested by @Abrams27
https://github.com/JetBrains/bazel-bsp/pull/24#discussion_r549535288

### LINKED ISSUE #29: buildTargetScalacOptions and buildTargetJavacOptions are returning empty
After working on extending the BSP testkit for the [extension methods](https://github.com/build-server-protocol/build-server-protocol/pull/154), I decided to test the sample repo used here against those methods. It would seem both `buildTargetScalacOptions` and `buildTargetJavacOptions` are returning empty responses. 

We should try and see what is going on there. Furthermore, after the testkit update is merged, we should take the chance to also make sure those requests are tested.

Abrams27:
Closes #27


andrefmrocha:
Thank you for this! Not sure if we should close the issue with this PR though, since it is also mentioned as we should update the tests with the updates that are coming to the testkit. Wdyt should be done? We could just open a new issue for that part or we could unlink this issue

Abrams27:
> Thank you for this! Not sure if we should close the issue with this PR though, since it is also mentioned as we should update the tests with the updates that are coming to the testkit. Wdyt should be done? We could just open a new issue for that part or we could unlink this issue

I think that we can unlink - this PR is mainly about a further refactor (I will link correct issues)

gerardd33:
> Thank you for this! Not sure if we should close the issue with this PR though, since it is also mentioned as we should update the tests with the updates that are coming to the testkit. Wdyt should be done? We could just open a new issue for that part or we could unlink this issue

I was going to resolve #29 in a separate PR but @Abrams27 seems to already have fixed it and the fix is smaller than we thought (also it is related to the rest of the work and structure refactor in this PR), so I would suggest letting this PR fix this and do the testkit part later in a separate PR. That is, not reverting these changes but also not closing #29 yet.

Abrams27:
Okey, I marked that it fixes #29, but we don't want to close it. Also, I decided to change only Utils in this PR (+ packages rearrange) to keep it smaller

src/main/java/org/jetbrains/bsp/bazel/commons/Constants.java
```diff
@@ -46,7 +46,6 @@
   public static final String TEST_RULE_TYPE = "test";
 
   public static final String BUILD_FILE_NAME = "BUILD";
-  public static final String WORKSPACE_FILE_NAME = "WORKSPACE";
```
agluszak, line 49: why was it moved?

src/main/java/org/jetbrains/bsp/bazel/server/bep/parsers/error/ErrorPositionParser.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  private static Position getPosition(String[] lineLocation) {
+    Integer lineNumber = getErrorLineNumber(lineLocation);
+    Integer characterIndex = getErrorCharacterIndex(lineLocation);
+
+    return new Position(lineNumber, characterIndex);
+  }
+
+  private static Integer getErrorLineNumber(String[] lineLocation) {
```
agluszak, line 45: isn't that unnecessary boxing and unboxing?

agluszak (commented):
2 minor nits, otherwise lgtm

src/main/java/org/jetbrains/bsp/bazel/commons/Constants.java
Abrams27: okey, this one could stay there - i will rollback it

src/main/java/org/jetbrains/bsp/bazel/server/bep/parsers/error/ErrorPositionParser.java
Abrams27: is there boxing and unboxing? i use `Integer` everywhere

agluszak (approved):

andrefmrocha (approved):
Looks good! Thanks!

src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
```diff
@@ -61,6 +61,19 @@ public BazelBspServerBuildManager(
     this.bazelRunner = bazelRunner;
   }
 
+  private static String getJavaVersion() {
```
gerardd33, line 64: I would put this method under ``getJVMBuildTarget`` where it's used.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/utils/ParsingUtils.java
```diff
@@ -0,0 +1,25 @@
...
+          Constants.TEST_RULE_TYPE, BuildTargetTag.TEST);
+  private static final String DEFAULT_RULE_TYPE = BuildTargetTag.NO_IDE;
+
+  public static String convertOutputToPath(String output, String prefix) {
+    String pathToFile = output.replaceAll("(//|:)", "/");
+    return prefix + pathToFile;
+  }
+
+  public static String getRuleType(String ruleClass) {
+    return RULE_CLASS_TO_RULE_TYPE.getOrDefault(ruleClass, DEFAULT_RULE_TYPE);
```
gerardd33, line 23: I would consider removing this class altogether and moving the contents of this class to ``BazelBspBuildManager`` (it's only used there) – it too is still pretty strongly related to the logic there, it's not much code and ``BazelBspBuildManager`` should still be split into several pieces in the future.

gerardd33 (commented):
Also 2 small notes, otherwise looks good.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/utils/ParsingUtils.java
Abrams27: this util is in this package and it is related with manager. Also, in general we do want to split logic, so I believe that it should stay in the separate class

Abrams27: + I rollbacked `getRuleType` method, because I missed `.contains`, what makes inconvenient to use map

gerardd33: Sure, just a suggestion. In this case maybe we should at least rename it to something that suggest these are parsing utils for the bsp package and not some general parsing utils? Don't have any good idea though.

Abrams27: it is in `.../bsp/utils` do we need specify it more?

gerardd33: ``BuildParsingUtils`` or ``BuildManagerParsingUtils`` would be fine for now but maybe we'd want to have here something from other places within the bsp package.

gerardd33: Actually maybe it is a good idea after all. There are not many such places that make sense.

Abrams27: done

gerardd33 (approved):
Good job!

PR #25, opened 2020-12-27
Implement buildTargetScalaMainClasses.

magda-aug:
This code should be refactored - significant part of the logic is copied from the buildTargetScalacOptions method. 

The code currently lacks tests.

PR #32, opened 2021-01-05
Solving problems with Scala- and Java-specific endpoints

gerardd33:
Closing #29 

Adding it already as a draft just so everyone knows the basic problem here is fixed and to discuss further work in #29.

I've added a class aggregating the implementation classes for use in ``startServer``. ``setLocalServices`` instead of ``setLocalService`` behaved strangely and froze causing him timeouts on the CI (maybe we misunderstood how that launcher works, if someone has an idea how to use it right, it would be nice to hear). For now it seems to be working again.

@andrefmrocha you mentioned you wanted to continue some more work in this issue including testing those endpoints. I'm not sure at this point how this fits with your testkit merge and with #25. I think we can talk about what exactly we want to do here and how during the Wednesday call?


src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -77,7 +81,7 @@ private void integrateBsp(BspIntegrationData bspIntegrationData) {
             .traceMessages(bspIntegrationData.getTraceWriter())
             .setOutput(bspIntegrationData.getStdout())
             .setInput(bspIntegrationData.getStdin())
-            .setLocalService(buildServer)
+            .setLocalService(aggregateBuildServerImpl)
             .setRemoteInterface(BuildClient.class)
             .setExecutorService(bspIntegrationData.getExecutor())
```
andrefmrocha, line 86: After some research I was able to find why the `setLocalServices` call probably didn't work. For a collection of local services, one should also set  a collection of remote interfaces using `setRemoteInterfaces`. This call can be seen in use [here](https://github.com/eclipse/lsp4j/blob/23f42a41288bb45ca088ca4cb75e2c9a70259e13/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/Launcher.java#L206). Lmk how that works out!

andrefmrocha (changes requested):
Got an idea on how to leave this services splitted, let's see how this works?

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
gerardd33: What exactly should I pass to ``setRemoteInterfaces`` then?

andrefmrocha: Disregard this :sweat: , I thought for a second that a client to support extension methods also needed to implement some classes which is not true

gerardd33:
I talked with Marcin and we decided it would be a good idea to merge this PR now with this fix and he will take care of testing the endpoints in a separate PR, since it is closely related to his other work. What do you think @andrefmrocha?

andrefmrocha:
> I talked with Marcin and we decided it would be a good idea to merge this PR now with this fix and he will take care of testing the endpoints in a separate PR, since it is closely related to his other work. What do you think @andrefmrocha?

Yeah that's okay, go ahead!

andrefmrocha (approved):

Abrams27 (approved):
LGTM, but I think that it would be nice to create a ticket for further `setLocalServices` investigation.

PR #33, opened 2021-01-12
[Post BazelBuilder refactor part 1] TargetsResolver update

Abrams27:
Introduces a new `TargetsResolver`!

This PR is all about:
- `TargetsResolver` -- now it is generic way to ask for build targets - java and scala implementations can use it
- `TargetsUtils` -- code contained around 6 occurrences of the same code
- `BuildClientLogger` -- does NOT throw exception any more and is moved closer to usages - now it's in `bep` package

After this PR:
`buildTargetResources` uses and`buildTargetScalaTestClasses`, `buildTargetScalaMainClasses` will use very similar logic -- probably this logic can be included in `TargetsResolver` or something similar should be created


src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsResolver.java
```diff
@@ -1,41 +1,175 @@
 package org.jetbrains.bsp.bazel.server.bsp.resolvers;
 
+import ch.epfl.scala.bsp4j.BuildTargetIdentifier;
+import com.google.common.collect.ImmutableList;
 import com.google.devtools.build.lib.query2.proto.proto2api.Build;
+import java.security.InvalidParameterException;
```
agluszak, line 6: Are you sure this is the exception you wanted?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/JavaBuildServerService.java
```diff
@@ -1,80 +1,43 @@
...
-  private final ActionGraphResolver actionGraphResolver;
-
-  public JavaBuildServerService(
-      BazelData bazelData,
-      TargetsResolver targetsResolver,
-      ActionGraphResolver actionGraphResolver) {
-    this.bazelData = bazelData;
-    this.targetsResolver = targetsResolver;
-    this.actionGraphResolver = actionGraphResolver;
+  private static final String JAVA_COMPILER_OPTIONS_NAME = "javacopts";
```
agluszak, line 17: we're moving these constants back and forth :P maybe let's keep them in Constants.java

agluszak (commented):
LGTM, minor nits

agluszak (commented):
LGTM, minor nits

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsResolver.java
Abrams27: `IllegalStateException` maybe?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/JavaBuildServerService.java
Abrams27: I feet that this is the proper final place for this constant

gerardd33 (commented):
Looks good, only two small things.

src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -49,7 +48,6 @@ public void startServer(BspIntegrationData bspIntegrationData) {
     BazelBspServerRequestHelpers serverRequestHelpers =
         new BazelBspServerRequestHelpers(serverLifetime);
 
-    TargetsResolver targetsResolver = new TargetsResolver(bazelRunner);
     ActionGraphResolver actionGraphResolver = new ActionGraphResolver(bazelRunner);
```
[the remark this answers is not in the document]
gerardd33: I mean the ``ActionGraphResolver``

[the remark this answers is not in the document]
Abrams27: right, forgot to remove

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -11,59 +10,42 @@
...
-      BazelData bazelData,
-      TargetsResolver targetsResolver,
-      ActionGraphResolver actionGraphResolver) {
-    this.bazelData = bazelData;
-    this.targetsResolver = targetsResolver;
-    this.actionGraphResolver = actionGraphResolver;
+  private final TargetsResolver<ScalacOptionsItem> targetsResolver;
+
+  public ScalaBuildServerService(BazelData bazelData, BazelRunner bazelRunner) {
+
```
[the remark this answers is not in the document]
Abrams27: after changes i forgot to remove this line

gerardd33 (approved):

agluszak (approved):

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsResolver.java
```diff
@@ -1,41 +1,174 @@
...
+  @FunctionalInterface
+  public interface ResultItemsCollector<T> {
+    T apply(
+        BuildTargetIdentifier target,
+        List<String> options,
+        List<String> classpath,
+        String classDirectory);
+  }
+
+  public static class Builder<T> {
```
andrefmrocha, line 124: Would it make sense for `TargetResolver` to live inside this class? To specificy how it pertains to this builder. Or maybe just having them in separate files. Wdyt?

andrefmrocha (commented):
Looks great, let just have a question, let me know what you think about it!

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsResolver.java
Abrams27: I'm not sure I understood the question correctly. `TargetsResolver` should (now it does) have a private constructor and then `Builder` is the only way to create instance of this class

andrefmrocha: But perhaps the `TargetResolver` class code can be inside the builder, that's what I meant

Abrams27: No, `TargetResolver` could exist without `Builder` but `Builder` cannot without `TargetResolver`. In general builder pattern looks like this - class + internal builder class

andrefmrocha: It might have been a bit too late for me to have answered, since I meant precisely the opposite of what I wrote :sweat_smile: . The Builder class code being inside the resolver. It's weird for me to see two public classes in a file without this pattern. I'm not against it though, if you disagree, you can go ahead and resolve this and merge :smile:

Abrams27: In general I don't like multiple classes in the same file too, but builder pattern is an exception. Maybe we should consider introducing a lombok project, which allows creating builder with one annotation.

andrefmrocha: We can consider doing that in the future, yes

PR #34, opened 2021-01-15
[Post BazelBuilder refactor part 2] TargetsRuleResolver

Abrams27:
Introducing ... `TargetsRuleResolver`!

It's a tiny change, but it'll be useful in `buildTargetScalaTestClasses` and `buildTargetScalaMainClasses` implementations.

Unfortunately, I tried, but I didn't manage to use it in `buildTargetResources` -- the logic is a bit too complicated, so I decided to leave it in the current state for a while. Nevertheless, I think that introduced mechanism could be extended and used there.

Also I thought that maybe it would be nice to introduce a [command pattern](https://en.wikipedia.org/wiki/Command_pattern) to this project -- api call = class. Logic, mostly, is very independent for every call (with few exceptions) so maybe it'd be nice to encapsulate it and get rid of ~ 400 lines classes. What do you think about it? If you think that it could be a good idea, I'll be able to implement it asap.

NOT RELATED WITH PR BUT I DECIDED TO FIX IT IN THIS PR:
- typo in tests `testsTopRun` -> `testsToRun`
- `ActionGraphParser` logic got simplified a bit -- unnecessary set creation (even intellij marked it)
- rename `TargetsResolver` -> `TargetsLanguageOptionsResolver` -- it's only used in `buildTargetJavacOptions` / `buildTargetScalacOptions`, but I'm still thinks about the best naming convention, any ideas?

### DIFF
--- a/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsLanguageOptionsResolver.java
+++ b/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsLanguageOptionsResolver.java
@@ -14,9 +14,10 @@ import org.jetbrains.bsp.bazel.server.bazel.data.BazelData;
 import org.jetbrains.bsp.bazel.server.bazel.data.BazelProcessResult;
 import org.jetbrains.bsp.bazel.server.bazel.params.BazelRunnerFlag;
 
-public class TargetsResolver<T> {
+public class TargetsLanguageOptionsResolver<T> {
 
   private static final List<String> ACTION_GRAPH_SUFFIXES = ImmutableList.of(".jar", ".js");
+
   private final BazelData bazelData;
   private final BazelRunner bazelRunner;
   private final ActionGraphResolver actionGraphResolver;
@@ -24,7 +25,7 @@ public class TargetsResolver<T> {
   private final List<String> languagesIds;
   private final ResultItemsCollector<T> resultItemsCollector;
 
-  private TargetsResolver(
+  private TargetsLanguageOptionsResolver(
       BazelData bazelData,
       BazelRunner bazelRunner,
       String compilerOptionsName,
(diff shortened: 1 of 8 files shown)


src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsLanguageOptionsResolver.java
```diff
@@ -14,17 +14,18 @@
 import org.jetbrains.bsp.bazel.server.bazel.data.BazelProcessResult;
 import org.jetbrains.bsp.bazel.server.bazel.params.BazelRunnerFlag;
 
-public class TargetsResolver<T> {
+public class TargetsLanguageOptionsResolver<T> {
```
agluszak, line 17: it's a bit of a mouthful, I think I prefer the previous name

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsRulesResolver.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.devtools.build.lib.query2.proto.proto2api.Build.Rule;
+import java.util.List;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import org.jetbrains.bsp.bazel.server.bazel.BazelRunner;
+import org.jetbrains.bsp.bazel.server.bazel.data.BazelProcessResult;
+import org.jetbrains.bsp.bazel.server.bazel.params.BazelRunnerFlag;
+
+public class TargetsRulesResolver<T> {
```
agluszak, line 14: maybe TargetRulesResolver? It seems more natural

agluszak: also I'm not sure what's the point of this class, it seems to be a very simple wrapper only

andrefmrocha: It is mentioned in the PR description as this will be useful for #22 and #25

andrefmrocha (commented):
Looks great, thanks! Let's just follow-up on the comments made by @agluszak

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsLanguageOptionsResolver.java
Abrams27: previous name was too general for sure -- now I introduced another targetsResolver

Abrams27:
I merge this PR -- it's minor change and it blocks my further changes. Any objections should be pointed in the following PR.

PR #35, opened 2021-01-16
buildTargetScalaTestClasses implementation 

Abrams27:
my implementation of `buildTargetScalaTestClasses` method

### DIFF
--- a/sample-repo/example/ExampleTest.scala
+++ b/sample-repo/example/ExampleTest.scala
@@ -1,7 +1,9 @@
+package example
+
 import dep.Dep
 
 object ExampleTest {
-  def test(): Unit = {
+  def main(args: Array[String]): Unit = {
     val s = "Sup" + Dep.list.head
     println(s)
   }
--- a/sample-repo/example/ExampleTest.scala
+++ b/sample-repo/example/ExampleTest.scala
@@ -1,0 +1,8 @@
+import dep.Dep
+
+object ExampleTest {
+  def test(): Unit = {
+    val s = "Sup" + Dep.list.head
+    println(s)
+  }
+}
--- a/src/main/java/org/jetbrains/bsp/bazel/server/bsp/impl/BuildServerImpl.java
+++ b/src/main/java/org/jetbrains/bsp/bazel/server/bsp/impl/BuildServerImpl.java
@@ -113,7 +113,6 @@ public class BuildServerImpl implements BuildServer {
 
   @Override
   public CompletableFuture<Object> workspaceReload() {
-    // TODO
-    return null;
+    return CompletableFuture.completedFuture(null);
   }
 }
--- a/src/main/java/org/jetbrains/bsp/bazel/server/bsp/impl/BuildServerImpl.java
+++ b/src/main/java/org/jetbrains/bsp/bazel/server/bsp/impl/BuildServerImpl.java
@@ -110,4 +110,10 @@ public class BuildServerImpl implements BuildServer {
     return serverRequestHelpers.executeCommand(
         () -> buildServerService.buildTargetCleanCache(cleanCacheParams));
   }
+
+  @Override
+  public CompletableFuture<Object> workspaceReload() {
+    // TODO
+    return null;
+  }
 }
--- a/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsUtils.java
+++ b/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsUtils.java
@@ -1,6 +1,7 @@
 package org.jetbrains.bsp.bazel.server.bsp.resolvers;
 
 import ch.epfl.scala.bsp4j.BuildTargetIdentifier;
+import com.google.devtools.build.lib.query2.proto.proto2api.Build;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -8,5 +9,19 @@ public final class TargetsUtils {
 
   public static List<String> getTargetsUris(List<BuildTargetIdentifier> targets) {
     return targets.stream().map(BuildTargetIdentifier::getUri).collect(Collectors.toList());
+  }
+
+  public static boolean doesRuleAttributesContain(Build.Rule rule, String attributeName) {
+    return rule.getAttributeList().stream()
+        .anyMatch(attribute -> isAttributeSpecifiedAndHasGivenName(attribute, attributeName));
+  }
+
+  public static boolean isAttributeSpecifiedAndHasGivenName(
+      Build.Attribute attribute, String name) {
+    boolean doesAttributeHaveGivenName = attribute.getName().equals(name);
+
+    return doesAttributeHaveGivenName
+        && attribute.hasExplicitlySpecified()
+        && attribute.getExplicitlySpecified();
   }
 }
--- a/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsUtils.java
+++ b/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsUtils.java
@@ -11,7 +11,7 @@ public final class TargetsUtils {
     return targets.stream().map(BuildTargetIdentifier::getUri).collect(Collectors.toList());
   }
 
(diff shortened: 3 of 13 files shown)


src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -39,19 +54,65 @@ public ScalaBuildServerService(BazelData bazelData, BazelRunner bazelRunner) {
...
+                    attribute, SCALA_TEST_SRCS_CLASSES_ATTRIBUTE_NAME))
+        .map(Attribute::getStringListValueList)
+        .flatMap(Collection::stream)
+        .collect(Collectors.toList());
+  }
+
+  private boolean doesAttributesContainTestClass(Build.Rule rule) {
+    return TargetsUtils.doesRuleAttributesContain(rule, SCALA_TEST_MAIN_CLASSES_ATTRIBUTE_NAME)
+        || TargetsUtils.doesRuleAttributesContain(rule, SCALA_TEST_SRCS_CLASSES_ATTRIBUTE_NAME);
   }
```
magda-aug, line 116: Shouldn't we check name of the rule in order to find test classes? ('scala_test' probably?)

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsUtils.java
```diff
@@ -9,4 +10,18 @@
   public static List<String> getTargetsUris(List<BuildTargetIdentifier> targets) {
     return targets.stream().map(BuildTargetIdentifier::getUri).collect(Collectors.toList());
   }
+
+  public static boolean doesRuleAttributesContain(Build.Rule rule, String attributeName) {
```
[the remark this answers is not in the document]
Abrams27: agree

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
Abrams27: right! I forgot to check it

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -1,32 +1,48 @@
...
 
 public class ScalaBuildServerService {
 
   private static final String SCALA_COMPILER_OPTIONS_NAME = "scalacopts";
   private static final List<String> SCALA_LANGUAGES_IDS =
       ImmutableList.of(Constants.SCALAC, Constants.JAVAC);
 
+  private static final String SCALA_TEST_RULE_NAME = "scala_test";
+  private static final String SCALA_TEST_MAIN_CLASSES_ATTRIBUTE_NAME = "main_class";
+  private static final String SCALA_TEST_SRCS_CLASSES_ATTRIBUTE_NAME = "srcs";
```
andrefmrocha, line 38: These attributes are generic to all `_test` and `_binary`. Maybe this should live in the `Constants`?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -39,19 +55,64 @@ public ScalaBuildServerService(BazelData bazelData, BazelRunner bazelRunner) {
...
+    List<String> classes = getTestClasses(rule);
+
+    return new ScalaTestClassesItem(target, classes);
+  }
+
+  private List<String> getTestClasses(Build.Rule rule) {
+    List<String> mainClasses = getTestMainClasses(rule);
+    List<String> srcsClasses = getTestSrcsClasses(rule);
+
+    return Stream.concat(mainClasses.stream(), srcsClasses.stream()).collect(Collectors.toList());
```
andrefmrocha, line 90: `main_class` is the defined name to the class where the main is, while `srcs` has the path to all the source files in a given rule. I'm not sure what you're trying to do here, but it doesn't seem correct. Can you pls explain?

andrefmrocha: More info on those values here: https://docs.bazel.build/versions/master/be/java.html#java_test

This is the docs for java, but all languages must abide by these rules

andrefmrocha (changes requested):
I have some doubts as to what is being done here. Also, we probably shouldn't merge this without tests right?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
Abrams27: I checked on more time and I agree that I misunderstood how `srcs` attribute works. But, now I found out that I should handle `test_class` attribute as well.

andrefmrocha: I wanted to check what would happen if both attributes were present and how to approach this and ended up finding out that `rules_scala` doesn't actually support `test_class`.

Abrams27: oh great, soo should I remove `srcsClasses` and connected logic and use only `main_class` attribute?

andrefmrocha: I do believe so!

andrefmrocha: Whenever they end up supporting more stuff that makes need to support `test_class`, we'll look into it. If that even happens :smile:

Abrams27: cool

andrefmrocha (approved):
Lgtm :rocket: !

andrefmrocha:
Are we handling tests in this PR or separately>

Abrams27:
> Are we handling tests in this PR or separately?

I will introduce tests in this PR, but firstly need to merge https://github.com/JetBrains/bazel-bsp/pull/32

agluszak (approved):

gerardd33 (approved):
Looks good!

src/main/java/org/jetbrains/bsp/bazel/server/bsp/impl/BuildServerImpl.java
```diff
@@ -110,4 +110,10 @@ public void onBuildExit() {
     return serverRequestHelpers.executeCommand(
         () -> buildServerService.buildTargetCleanCache(cleanCacheParams));
   }
+
+  @Override
+  public CompletableFuture<Object> workspaceReload() {
+    // TODO
+    return null;
```
andrefmrocha, line 117: It's probably best to return a [Completed Future](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completedFuture-U-) here

sample-repo/example/ExampleTest.scala
```diff
@@ -0,0 +1,8 @@
+import dep.Dep
+
+object ExampleTest {
+  def test(): Unit = {
```
andrefmrocha, line 4: To keep this consistent, this object should have a main method, since this is required by main_class

andrefmrocha (changes requested):
Just some minor nitpicks

andrefmrocha (approved):
Lgtm :rocket: !

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
```diff
@@ -15,6 +15,9 @@
 
   private static final Logger LOGGER = LogManager.getLogger(BazelBspServerTest.class);
 
+  private static final Integer SUCCES_EXIT_CODE = 0;
```
jastice, line 18: typo: success

jastice (approved):
LGTM!

PR #36, opened 2021-01-20
Implement buildTargetScalaMainClasses

magda-aug:


src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -8,25 +11,33 @@
...
+import org.jetbrains.bsp.bazel.server.bsp.resolvers.TargetRulesResolver;
 import org.jetbrains.bsp.bazel.server.bsp.resolvers.TargetsLanguageOptionsResolver;
 
 public class ScalaBuildServerService {
 
   private static final String SCALA_COMPILER_OPTIONS_NAME = "scalacopts";
   private static final List<String> SCALA_LANGUAGES_IDS =
       ImmutableList.of(Constants.SCALAC, Constants.JAVAC);
+  private static final String MAIN_CLASS_ATTR_NAME = "main_class";
+  private static final String ARGS_ATTR_NAME = "args";
```
andrefmrocha, line 34: As I had mentioned in #35 in [a comment](https://github.com/JetBrains/bazel-bsp/pull/35#discussion_r559807761), these values are shared. Perhaps they should live in the Constants file

andrefmrocha (commented):
Looks great! Just a minor detail pointed out. Looking forward to seeing those tests!

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -54,10 +63,55 @@ public ScalaBuildServerService(BazelData bazelData, BazelRunner bazelRunner) {
     return CompletableFuture.completedFuture(new ScalaTestClassesResult(new ArrayList<>()));
   }
 
-  public CompletableFuture<ScalaMainClassesResult> buildTargetScalaMainClasses(
+  public Either<ResponseError, ScalaMainClassesResult> buildTargetScalaMainClasses(
       ScalaMainClassesParams scalaMainClassesParams) {
-    System.out.printf("DWH: Got buildTargetScalaMainClasses: %s%n", scalaMainClassesParams);
-    // TODO(illicitonion): Populate
-    return CompletableFuture.completedFuture(new ScalaMainClassesResult(new ArrayList<>()));
+    TargetRulesResolver<ScalaMainClassesItem> targetRulesResolver =
```
Abrams27, line 68: it doesn't have to be created on every method call - i think that it should be moved to the constructor

Abrams27 (commented):
looks great! Good job, I left only one minor comment. Also, I have one question - you are not filtering rules out, it means that you will return main classes even for `scala_test` rules, is it okey or not?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -54,10 +63,55 @@ public ScalaBuildServerService(BazelData bazelData, BazelRunner bazelRunner) {
...
+        .collect(Collectors.toList());
+  }
+
+  private Stream<Build.Attribute> getAttribute(Build.Rule rule, String name) {
+    return rule.getAttributeList().stream()
+        .filter(
+            attr ->
+                attr.getName().equals(name)
+                    && attr.hasExplicitlySpecified()
+                    && attr.getExplicitlySpecified());
```
gerardd33, line 115: Can be extracted, this is done in Marcin's PR.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -54,10 +63,55 @@ public ScalaBuildServerService(BazelData bazelData, BazelRunner bazelRunner) {
...
+    BuildTargetIdentifier targetId = new BuildTargetIdentifier(rule.getName());
+    List<ScalaMainClass> mainClasses = collectMainClasses(rule);
+    return new ScalaMainClassesItem(targetId, mainClasses);
+  }
+
+  private List<ScalaMainClass> collectMainClasses(Build.Rule rule) {
+    List<String> targetOptions =
+        getAttribute(rule, SCALA_COMPILER_OPTIONS_NAME)
+            .flatMap(attr -> attr.getStringListValueList().stream())
+            .collect(Collectors.toList());
```
gerardd33, line 89: Isn't it the same case as the two methods extracted below, that is, could it be extracted to something like _getTargetOptions_?

gerardd33 (commented):
Looks great! We just should make sure it's sensibly rebased on Marcin's changes in the same files (after merging them), especially making sure it's consistent. I marked one such place below but make sure that in general analogous things are done the same way in both PRs, extracting parts to methods on the same levels, using common helper methods etc.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/BspImplementationHub.java
```diff
@@ -70,6 +70,12 @@ public void onBuildExit() {
     return buildServer.workspaceBuildTargets();
   }
 
+  @Override
+  public CompletableFuture<Object> workspaceReload() {
+    // TODO
+    return null;
```
andrefmrocha, line 76: This should probably call the `BuildServerImpl` one, just for consistency sake

src/main/java/org/jetbrains/bsp/bazel/server/bsp/impl/BuildServerImpl.java
```diff
@@ -61,6 +61,12 @@ public void onBuildExit() {
     return buildServerService.workspaceBuildTargets();
   }
 
+  @Override
+  public CompletableFuture<Object> workspaceReload() {
+    // TODO
+    return null;
```
andrefmrocha, line 67: It's probably best to return a [Completed Future](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#completedFuture-U-) to make sure this does not break on the client side.

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTestData.java
```diff
@@ -104,4 +109,16 @@
...
+      new ScalaMainClassesParams(Collections.singletonList(ID_1));
+
+  static final ScalaMainClassesResult EXPECTED_SCALA_MAIN_CLASSES =
+      new ScalaMainClassesResult(
+          Collections.singletonList(
+              new ScalaMainClassesItem(
+                  ID_1,
+                  Collections.singletonList(
+                      new ScalaMainClass(
+                          "example.Example", Collections.emptyList(), Collections.emptyList())))));
```
andrefmrocha, line 123: We probably should guarantee that the rest of the Scala Main Class is also working, by providing arguments and jvmOptions to the example class, wdyt?

andrefmrocha (changes requested):
Looks great! Just have some suggestions to make this even more robust

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
magda-aug: It's not exactly the same, two methods use getStringListValueList (because the attributes are of type string list) and one uses getStringValue. 
Anyway, I extracted the methods with the same logic

magda-aug:
> looks great! Good job, I left only one minor comment. Also, I have one question - you are not filtering rules out, it means that you will return main classes even for `scala_test` rules, is it okey or not?

The specification says "The build target main classes request is sent from the client to the server to query for the list of main classes that can be fed as arguments to buildTarget/run" so I would guess it is ok to return test classes, but I am not sure...
@andrefmrocha what do you think about this?

andrefmrocha:
> The specification says "The build target main classes request is sent from the client to the server to query for the list of main classes that can be fed as arguments to buildTarget/run" so I would guess it is ok to return test classes, but I am not sure...
> @andrefmrocha what do you think about this?

Perhaps the semantics is not ideal here, but one would assume these are main classes that refer to production level of main classes, and not tests, since there doesn't seem to be any good reason to be able to run a test as a normal binary, without providing the interface existing for tests. Pinging @jastice just to make sure

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTestData.java
andrefmrocha: You should have access to a `scalacopts` and I think `jvm_flags` should be substituted by a union of `jvm_flags`, `scalac_jvm_flags` and `javac_jvm_flags`: https://github.com/bazelbuild/rules_scala/blob/773c992c0fcbf0e18355eaef1783c49915455c12/docs/scala_binary.md .

src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/ScalaBuildServerService.java
```diff
@@ -54,10 +67,52 @@ public ScalaBuildServerService(BazelData bazelData, BazelRunner bazelRunner) {
...
+  private ScalaMainClassesItem mapRuleToMainClassesItem(Build.Rule rule) {
+    BuildTargetIdentifier targetId = new BuildTargetIdentifier(rule.getName());
+    List<ScalaMainClass> mainClasses = collectMainClasses(rule);
+    return new ScalaMainClassesItem(targetId, mainClasses);
+  }
+
+  private List<ScalaMainClass> collectMainClasses(Build.Rule rule) {
+    Set<String> targetOptionsSet =
+        new LinkedHashSet<>(
+            collectAttributesFromStringListValues(rule, SCALA_COMPILER_OPTIONS_ATTR_NAME));
```
andrefmrocha, line 90: I think I may have induced you in error with this, so I'm sorry about that :sweat_smile: . The `MainClass` request only receives options from the jvm, meaning the `jvm_flags` . The `scalacopts` field should be use on the `buildTarget/ScalacOptions`, not here

andrefmrocha (changes requested):
There's a problem here that's mostly miscommunication, I'm sorry about that :sweat_smile:

jastice:
> > The specification says "The build target main classes request is sent from the client to the server to query for the list of main classes that can be fed as arguments to buildTarget/run" so I would guess it is ok to return test classes, but I am not sure...
> > @andrefmrocha what do you think about this?
> 
> Perhaps the semantics is not ideal here, but one would assume these are main classes that refer to production level of main classes, and not tests, since there doesn't seem to be any good reason to be able to run a test as a normal binary, without providing the interface existing for tests. Pinging @jastice just to make sure

Yeah, I would only return classes that can be run as standalone application, i.e. have a main method. They can be within test targets, of course.

jastice (approved):

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTestData.java
```diff
@@ -110,6 +115,20 @@
   static final InverseSourcesResult EXPECTED_INVERSE_SOURCES =
       new InverseSourcesResult(ImmutableList.of(ID_2));
 
+  static final ScalaMainClassesParams SCALA_MAIN_CLASSES_PARAMS =
+      new ScalaMainClassesParams(ImmutableList.of(ID_1));
+
+  static final ScalaMainClassesResult EXPECTED_SCALA_MAIN_CLASSES =
```
Abrams27, line 121: I think that it would be nice to add more targets here - with empty params and opts or without `main_class`

Abrams27 (commented):
great job! only one comment about test

Abrams27 (approved):
looks great!

andrefmrocha (approved):
Looks great! Awesome work!

PR #37, opened 2021-01-30
javacoptions and scalacoptions tests

Abrams27:
Tests for `buildTargetScalacOptions` and `buildTargetJavacOptions` methods.

Tests only `target` and `options`, since `classpath` and `classDirectory` look very random (depend on local directories structure), or should it be somehow addressed?


andrefmrocha:
> Tests only `target` and `options`, since `classpath` and `classDirectory` look very random (depend on local directories structure), or should it be somehow addressed?

I think the testkit is, to some extent, already prepared for most of randomness since it will only use a `.contains()` call to address that fact instead of an `.equals()`. At least the classpath is very important to test, to make sure response contains the scala-compiler, scala-library and scala-reflect.

andrefmrocha:
I've been doing some investigation regarding this `Protocol message had invalid UTF-8` error message the CI is providing and one thing I found out is that this should be, in some way, bazel related, as switching back to an older version of bazel fixes this.

Might you check the version of bazel you're running locally @Abrams27? With bazel 4.0.0, I had this error.

andrefmrocha:
After some more research, I've found out that the protobuf output of aquery has suffered breaking changes, which means we will have so support 2 different ways of obtaining the dependency graph and parsing it. More info on that here: https://github.com/bazelbuild/bazel/issues/10358

Abrams27:
> After some more research, I've found out that the protobuf output of aquery has suffered breaking changes, which means we will have so support 2 different ways of obtaining the dependency graph and parsing it. More info on that here: [bazelbuild/bazel#10358](https://github.com/bazelbuild/bazel/issues/10358)

Ohhh, does it mean implementing a new parser?

andrefmrocha:
> > After some more research, I've found out that the protobuf output of aquery has suffered breaking changes, which means we will have so support 2 different ways of obtaining the dependency graph and parsing it. More info on that here: [bazelbuild/bazel#10358](https://github.com/bazelbuild/bazel/issues/10358)
> 
> Ohhh, does it mean implementing a new parser?

Yes it does, otherwise this only works for bazel 4.0 downwards, excludingly. But I've already implemented it for you guys, since this is a breaking issue. We'll discuss it tomorrow!

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTestData.java
```diff
@@ -119,41 +129,88 @@
...
+                  "__main__/bazel-out/darwin-fastbuild/bin/example/"),
+              new ScalacOptionsItem(
+                  DEP_DEP_TARGET,
+                  ImmutableList.of(),
+                  ImmutableList.of(
+                      "__main__/external/io_bazel_rules_scala_scala_library/scala-library-2.12.8.jar",
+                      "__main__/external/io_bazel_rules_scala_scala_reflect/scala-reflect-2.12.8.jar",
+                      "__main__/bazel-out/darwin-fastbuild/bin/external/io_bazel_rules_scala/src/java/io/bazel/rulesscala/scalac/scalac.jar",
+                      "__main__/bazel-out/darwin-fastbuild/bin/dep/deeper/deeper.jar"),
+                  "__main__/bazel-out/darwin-fastbuild/bin/dep/")));
```
andrefmrocha, line 215: Pls remove the `__main__/bazel-out/darwin-fastbuild` from the beggining of the paths. `darwin-fastbuild` is a path that's specifically created in OSX. Other OS's use different names :smile:

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
```diff
@@ -83,7 +83,19 @@ public void run() {
...
+            () ->
+                client.testJavacOptions(
+                    BazelBspServerTestData.JAVAC_OPTIONS_PARAMS,
+                    BazelBspServerTestData.EXPECTED_JAVAC_OPTIONS)),
+        new BazelBspServerSingleTest(
+            "scalacopts test",
+            () ->
+                client.testScalacOptions(
+                    BazelBspServerTestData.SCALAC_OPTIONS_PARAMS,
+                    BazelBspServerTestData.EXPECTED_SCALAC_OPTIONS))
```
andrefmrocha, line 98: They should be commented out as well, since they are not passing yet. Wdyt?

andrefmrocha (changes requested):
As discussed during our meeting, here's a few changes that need to be done!

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
andrefmrocha: I just remembered that it makes more sense for us to just specify the version of bazel of the sample repository as one prior to 4.0.0 right off the bat :sweat_smile:

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTestData.java
Abrams27: ohh righ

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
Abrams27: I decided to set the version on `3.7.2` (most recent before `4.0.0`)

andrefmrocha (approved):
Looks great! :rocket:

PR #38, opened 2021-02-03
workspaceReload implementation

Abrams27:


src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
```diff
@@ -335,4 +335,12 @@ private SourcesItem createSourcesForLabelAndItemsAndRoots(
     }
     return Either.forRight(result);
   }
+
+  public Either<ResponseError, Object> workspaceReload() {
+    bazelRunner.commandBuilder()
+        .sync()
+        .executeBazelCommand();
```
andrefmrocha, line 342: Not sure if this is what we want here. Sync all by itself will unconditionally execute all the rules in the `WORKSPACE` file as if all repositories are out of date. This is, naturally, a very costly way of guaranteeing that everything is up to date... 

After doing some more research I found the `bazel fetch` command. I think is more more appropriate to our use case! Check it out [here](https://docs.bazel.build/versions/master/command-line-reference.html#commands)

Abrams27: should we consider doing something more or `bazel fetch` call is enough?

andrefmrocha: Given the definition that was provided by @jastice, I think guaranteeing that the current dependencies are fetched and updated should be enough. Can you confirm justin?

jastice: Yes I think that is sufficient

andrefmrocha (approved):
Looks great! :rocket:

PR #41, opened 2021-02-17
buildClient is now used in bazel runner

Abrams27:
I had to change a bit packaging and bazel runner creation (I don't like this setter as well, but entire initialization is a bit messy).
I can see some logs now but I cannot test it fully because of the lock bug.


andrefmrocha (approved):
> I had to change a bit packaging and bazel runner creation (I don't like this setter as well, but entire initialization is a bit messy).

I'm not sure if we can really escape this initialization, however, due to the circular dependency that is, in fact, introduces. I think this looks great nonetheless, thanks!

gerardd33 (approved):
Looks good! The situation with the constructing seems to be the same as with _BepServer_ and I don't think there's a better way of doing this and getting rid of the setters due to that complex circular initialisation. But it's still not that bad and reduced to minimum.

PR #42, opened 2021-02-25
Action graph parser v2

andrefmrocha:
Add support for the new action graph that all versions from bazel 4.0.0 onwards will be using to represent their data whilst still keeping the backwards compatability with the previous action graph


src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,49 @@
...
+public class SemanticVersion implements Comparable<SemanticVersion> {
+
+  private final String version;
+
+  public final String get() {
+    return this.version;
+  }
+
+  public SemanticVersion(String version) {
+    if (version == null) throw new IllegalArgumentException("Version can not be null");
```
Abrams27, line 12: I think that maybe it would be better to stick with adding brackets around every statement

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,49 @@
...
+    if (!version.matches("[0-9]+(\\.[0-9]+)*"))
```
Abrams27, line 13: I believe that it would be better to extract this regex to the variable with proper name

Abrams27: and again brackets

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,49 @@
...
+      throw new IllegalArgumentException("Invalid version format");
+    this.version = version;
+  }
+
+  static SemanticVersion fromReleaseData(String version) {
+    return new SemanticVersion(version.split(" ")[1]);
+  }
+
+  @Override
+  public int compareTo(SemanticVersion other) {
```
Abrams27, line 23: As i understand versioning format correctly - it is `xx.xx`, right? Maybe we should explicitly extract first part and second (maybe, getMajorVersion, getMinorVersion, we can even keep it in this format) and then in comparison explicitly compare majors first and then minors

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/ActionGraphResolver.java
```diff
@@ -26,9 +34,18 @@ public ActionGraphParser getActionGraphParser(List<String> targets, List<String>
...
               .executeBazelCommand();
 
-      ActionGraphContainer actionGraphContainer =
-          AnalysisProtos.ActionGraphContainer.parseFrom(process.getStdoutStream());
-      return new ActionGraphParser(actionGraphContainer);
+      ActionGraphParser actionGraphParser;
+      if (bazelData.getVersion().compareTo(ACTION_GRAPH_V2_VERSION) < 0) {
+        AnalysisProtos.ActionGraphContainer actionGraphContainer =
+            AnalysisProtos.ActionGraphContainer.parseFrom(process.getStdoutStream());
+        actionGraphParser = new ActionGraphV1Parser(actionGraphContainer);
```
Abrams27, line 41: return statement can be here

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
```diff
@@ -11,15 +11,15 @@
...
 
   private final AnalysisProtos.ActionGraphContainer actionGraph;
 
-  public ActionGraphParser(AnalysisProtos.ActionGraphContainer actionGraph) {
+  public ActionGraphV1Parser(AnalysisProtos.ActionGraphContainer actionGraph) {
     this.actionGraph = actionGraph;
   }
 
-  public List<String> getInputs(String target, List<String> suffixes) {
+  private List<String> getInputs(String target, List<String> suffixes) {
```
Abrams27, line 22: I think that this method should be under the `getInputsAsUri` method

Abrams27: + it would be nice to add `@Override` annotation for `getInputsAsUri` and `getOutputs`

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,111 @@
...
+  private final Map<Integer, AnalysisProtosV2.PathFragment> pathFragmentMap;
+
+  public ActionGraphV2Parser(AnalysisProtosV2.ActionGraphContainer actionGraph) {
+    this.actionGraph = actionGraph;
+    pathFragmentMap =
+        actionGraph.getPathFragmentsList().stream()
+            .collect(Collectors.toMap(AnalysisProtosV2.PathFragment::getId, fragment -> fragment));
+  }
+
+  private List<String> getInputs(String target, List<String> suffixes) {
```
Abrams27, line 20: like previous - should be under the it's call

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
```diff
@@ -18,29 +19,52 @@
...
-    List<BazelBspServerSingleTest> testsTopRun = getTestsToRun();
-    runTests(testsTopRun);
+    List<BazelBspServerSingleTest> testsToRun =
+        Stream.concat(getSampleRepoTests().stream(), getActionGraphV2Tests().stream())
+            .collect(Collectors.toList());
+    runTests(testsToRun);
   }
 
-  private List<BazelBspServerSingleTest> getTestsToRun() {
+  private List<BazelBspServerSingleTest> getActionGraphV2Tests() {
```
Abrams27, line 39: great idea, maybe we ActionGraphV1 test should be extracted to the different scope as well?

Abrams27: i know that it would be in the same repo, but I think that it would be more flexible

Abrams27 (changes requested):
really great job! left few comments

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
andrefmrocha: Semantic versioning usually divides into three versions in a `xx.xx.xx` format. But I can do that either way!

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
andrefmrocha: Ah so you mean just creating a different listing for those tests? sure

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/ActionGraphResolver.java
andrefmrocha: I'm not sure I understand. You prefer for me to have two different returns for each if branch?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+        .collect(Collectors.toList());
+  }
+
+  private List<String> getInputs(String target, List<String> suffixes) {
+    return getActions(target).stream()
+        .flatMap(action -> action.getInputDepSetIdsList().stream())
+        .flatMap(
+            depset -> {
+              Queue<Integer> queue = new ArrayDeque<>();
+              queue.add(depset);
```
agluszak, line 33: it appears that you create a new queue and always add only a single element to it

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private List<AnalysisProtosV2.Action> getActions(String targetLabel) {
+    int targetId = getTargetId(targetLabel);
+    return actionGraph.getActionsList().stream()
+        .filter(action -> targetId == action.getTargetId())
+        .collect(Collectors.toList());
+  }
+
+  private List<AnalysisProtosV2.Artifact> expandDepsetToArtifacts(Queue<Integer> idsToExpand) {
```
agluszak, line 87: as I mentioned above, this queue appears to always have exactly one element

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+    HashSet<Integer> expandedIds = new HashSet<>();
+
+    HashSet<Integer> artifactIds = new HashSet<>();
+    while (!idsToExpand.isEmpty()) {
+      Integer depsetId = idsToExpand.remove();
+      if (expandedIds.contains(depsetId)) {
```
agluszak, line 93: I understand that your intent here is to remove the duplicates from the queue?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+        .flatMap(
+            depset -> {
+              Queue<Integer> queue = new ArrayDeque<>();
+              queue.add(depset);
+              return expandDepsetToArtifacts(queue).stream();
+            })
+        .map(AnalysisProtosV2.Artifact::getPathFragmentId)
+        .map(pathFragmentId -> "exec-root://" + constructPath(pathFragmentId))
+        .filter(path -> suffixes.stream().anyMatch(path::endsWith))
+        .collect(Collectors.toCollection(TreeSet::new))
```
agluszak, line 39: I think you can use `disctinct()` here instead

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+        .stream()
+        .collect(Collectors.toList());
+  }
+
+  private String constructPath(final Integer pathFragmentId) {
```
agluszak, line 44: why is this parameter final?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+    int currId = pathFragmentId;
+    Stack<String> pathStack = new Stack<>();
+    while (currId != 0) {
+      AnalysisProtosV2.PathFragment fragment = pathFragmentMap.get(currId);
+      pathStack.add(fragment.getLabel());
+      currId = fragment.getParentId();
+    }
+    final StringBuilder path = new StringBuilder();
```
agluszak, line 52: why is this variable final?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+    while (pathStack.size() != 1) path.append(pathStack.pop()).append("/");
```
agluszak, line 53: imo the logic here would be simpler if you collected the path fragments into a list, reversed the list, and then joined its contents with "/" instead of using a stack

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+import com.google.devtools.build.lib.analysis.AnalysisProtosV2;
+import java.util.*;
+import java.util.stream.Collectors;
+import org.jetbrains.bsp.bazel.commons.Uri;
+
+public class ActionGraphV2Parser implements ActionGraphParser {
+  private final AnalysisProtosV2.ActionGraphContainer actionGraph;
+  private final Map<Integer, AnalysisProtosV2.PathFragment> pathFragmentMap;
+
+  public ActionGraphV2Parser(AnalysisProtosV2.ActionGraphContainer actionGraph) {
```
agluszak, line 13: could you add a link to the specification of what we're actually parsing here?

agluszak: also I think I could receive some unit testing love

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+            .collect(Collectors.toSet());
+    return actionGraph.getArtifactsList().stream()
+        .filter(artifact -> artifactIds.contains(artifact.getId()))
+        .map(AnalysisProtosV2.Artifact::getPathFragmentId)
+        .map(this::constructPath)
+        .filter(path -> suffixes.stream().anyMatch(path::endsWith))
+        .collect(Collectors.toList());
+  }
+
+  private int getTargetId(String needle) {
```
agluszak, line 72: I'm not sure if "needle" is the best name for the parameter. I don't quite understand what this method is supposed to do - if it fails to find the the target it returns some default value

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+    return getActions(target).stream()
+        .flatMap(action -> action.getInputDepSetIdsList().stream())
+        .flatMap(
+            depset -> {
+              Queue<Integer> queue = new ArrayDeque<>();
+              queue.add(depset);
+              return expandDepsetToArtifacts(queue).stream();
+            })
+        .map(AnalysisProtosV2.Artifact::getPathFragmentId)
+        .map(pathFragmentId -> "exec-root://" + constructPath(pathFragmentId))
```
agluszak, line 37: "exec-root://" can be extracted as a constant

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
```diff
@@ -18,29 +19,82 @@
...
   private static final Integer FAIL_EXIT_CODE = 1;
 
-  private final TestClient client;
   private final ExecutorService executorService = Executors.newCachedThreadPool();
 
   public BazelBspServerTest() {
     LOGGER.info("Creating TestClient...");
 
-    this.client =
+    LOGGER.info("Created TestClient");
```
agluszak, line 27: it looks like nothing happens between these two messages now

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,79 @@
...
+    if (parts.length != 3) {
+      throw new IllegalArgumentException("Invalid number of version parts");
+    }
+
+    this.version = version;
+    this.versionParts = parts;
+  }
+
+  public final int getMajorVersion() {
+    return Integer.parseInt(versionParts[MAJOR_VERSION_LOCATION]);
```
agluszak, line 31: it would be better to do the parsing in the constructor and simply return the value here

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+
+  public ActionGraphV2Parser(AnalysisProtosV2.ActionGraphContainer actionGraph) {
+    this.actionGraph = actionGraph;
+    pathFragmentMap =
+        actionGraph.getPathFragmentsList().stream()
+            .collect(Collectors.toMap(AnalysisProtosV2.PathFragment::getId, fragment -> fragment));
+  }
+
+  @Override
+  public List<String> getInputsAsUri(String target, String execRoot) {
```
agluszak, line 21: this method looks the same for both versions, am I right?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
...
+    return getInputs(target, Lists.newArrayList(".jar", "js")).stream()
+        .map(exec_path -> Uri.fromExecPath(exec_path, execRoot).toString())
```
agluszak, line 23: exec_path should be execPath

agluszak (changes requested):

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,109 @@
+package org.jetbrains.bsp.bazel.server.bsp.resolvers.actiongraph;
+
+import com.google.common.collect.Lists;
+import com.google.devtools.build.lib.analysis.AnalysisProtosV2;
+import java.util.*;
```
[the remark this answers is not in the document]
andrefmrocha: Should I remove it?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: I believe that's the case but I'm not 100% since this logic is more or less duplicate of the first action graph parser, which I did not build

andrefmrocha: Though the code is not easy to share since this changed from string to integer

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: This logic is shared with the first action graph parser which I did not build. I think if we want to play around with that and make changes perhaps it should be in another PR? I'm not entirely aware of what's going on in these functions. But from what I understand the goal is to provide an empty target to generate an empty classpath list, not sure if I'm right though.

Regarding the variable name, should I change both of them then?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: Will change in both parsers

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: > could you add a link to the specification of what we're actually parsing here?

If you have the protocol buffers plugin it already redirects you to the .proto file of the implementation. Do you still think it makes sense?

andrefmrocha: > also I think I could receive some unit testing love

You mean we should have some unit tests for the action graph parsers? How would we go about doing those? Would we add some example binaries to the resources?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/ActionGraphResolver.java
Abrams27: yes, assigning to `actionGraphParser` is redundant

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
Abrams27: Well, I think that some logic should be extracted from both parsers to the separate class or to the interface

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: My mistake, will remove

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: My mistake

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: Sure!

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: Extracted to interface

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: Will add the queue creation to the function body

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: The goal here is not to expand in the same nodes again in the dependency graph

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
andrefmrocha: Can I just remove them then?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,79 @@
+package org.jetbrains.bsp.bazel.server.bazel.data;
+
+public class SemanticVersion implements Comparable<SemanticVersion> {
+
+  private static final String SEMANTIC_VERSION_REGEX = "[0-9]+(\\.[0-9]+)*";
+  private static final int MAJOR_VERSION_LOCATION = 0;
+  private static final int MINOR_VERSION_LOCATION = 1;
+  private static final int PATCH_VERSION_LOCATION = 1;
```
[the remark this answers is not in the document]
andrefmrocha: Yap, thanks!

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,79 @@
...
+    return Integer.parseInt(versionParts[PATCH_VERSION_LOCATION]);
+  }
+
+  static SemanticVersion fromReleaseData(String version) {
+    return new SemanticVersion(version.split(" ")[1]);
+  }
+
+  @Override
+  public int compareTo(SemanticVersion other) {
+    if (other == null) return 1;
```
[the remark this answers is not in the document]
andrefmrocha: Thanks, didnt know this was a thing :smile:

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public final int getMinorVersion() {
+    return Integer.parseInt(versionParts[MINOR_VERSION_LOCATION]);
+  }
+
+  public final int getPatchVersion() {
+    return Integer.parseInt(versionParts[PATCH_VERSION_LOCATION]);
+  }
+
+  static SemanticVersion fromReleaseData(String version) {
```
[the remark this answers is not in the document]
andrefmrocha: My mistake, fixed!

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return version.hashCode();
+  }
+
+  @Override
+  public boolean equals(Object that) {
+    if (this == that) return true;
```
Abrams27, line 69: should we stick with brackets around every statement?

gerardd33 (changes requested):
Really good job! It seems others have already marked the important places, so I'm leaving only a few stylistic comments and looking at how those other places resolve.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
```diff
@@ -67,7 +49,14 @@ private String getTargetId(String needle) {
         .collect(Collectors.toList());
   }
 
-  private List<AnalysisProtos.Artifact> expandDepsetToArtifacts(Queue<String> idsToExpand) {
+  private List<AnalysisProtos.Artifact> expandDepsetToArtifacts(String idToExpand) {
+    Queue<String> idsToExpand =
```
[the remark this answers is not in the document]
andrefmrocha: Can an interable be used if you're updating the queue whilst iterating through it?

Abrams27, line 53: I'm not sure what is happening here, I know that this is old code, but it might be updated a bit

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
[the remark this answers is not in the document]
gerardd33: Yeah, IntelliJ should then automatically give you the specific narrower imports.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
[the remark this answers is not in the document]
gerardd33: I meant just creating it using this method, I thought it could be used like [this](https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Lists.html#newArrayList(E...)), not sure why there's no analogous method here, but still I'm not sure what this

```
{
            add(idToExpand);
          }
}
```

does exactly? Why can't it be a simple one-ine initialisation?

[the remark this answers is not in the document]
andrefmrocha: Technically this is just a way of defining a series of operations that are related to the initialization. It's more semantics than anything. But I can use a list, sure.

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
```diff
@@ -18,29 +19,76 @@
   private static final Integer SUCCESS_EXIT_CODE = 0;
   private static final Integer FAIL_EXIT_CODE = 1;
 
-  private final TestClient client;
   private final ExecutorService executorService = Executors.newCachedThreadPool();
 
-  public BazelBspServerTest() {
-    LOGGER.info("Creating TestClient...");
+  @SafeVarargs
+  static <T> Stream<T> concat(Stream<T>... streams) {
```
Abrams27, line 25: why it has a default visibility?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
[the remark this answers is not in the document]
andrefmrocha: I dont mind doing that but I'm thinking if it doesn't make sense for us to guarantee this by a styling rule as @agluszak said?

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,74 @@
...
+  public SemanticVersion(String version) {
+    if (version == null) {
+      throw new IllegalArgumentException("Version can not be null");
+    }
+    if (!version.matches(SEMANTIC_VERSION_REGEX)) {
+      throw new IllegalArgumentException("Invalid version format");
+    }
+
+    String[] parts = version.split("\\.");
+    if (parts.length != 3) {
```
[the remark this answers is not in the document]
andrefmrocha: Good call, thanks!

src/main/java/org/jetbrains/bsp/bazel/server/bazel/data/SemanticVersion.java
```diff
@@ -0,0 +1,74 @@
...
+    this.majorVersion = Integer.parseInt(parts[MAJOR_VERSION_LOCATION]);
+    this.minorVersion = Integer.parseInt(parts[MINOR_VERSION_LOCATION]);
+    this.patchVersion = Integer.parseInt(parts[PATCH_VERSION_LOCATION]);
+  }
+
+  public static SemanticVersion fromReleaseData(String version) {
+    return new SemanticVersion(version.split(" ")[1]);
+  }
+
+  public final int getMajorVersion() {
```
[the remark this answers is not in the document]
andrefmrocha: My mistake

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
andrefmrocha: I've tried to streamify it a bit, let me know if there's anything else you think should be done!

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
```diff
@@ -67,7 +50,9 @@ private String getTargetId(String needle) {
         .collect(Collectors.toList());
   }
 
-  private List<AnalysisProtos.Artifact> expandDepsetToArtifacts(Queue<String> idsToExpand) {
+  private List<AnalysisProtos.Artifact> expandDepsetToArtifacts(String idToExpand) {
+    Queue<String> idsToExpand = new ArrayDeque<>(Lists.newArrayList(idToExpand));
```
Abrams27, line 54: well, this queue will contain only one element - it can be replaced with `Optional` and then execute filtering below

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -68,12 +69,7 @@ private int getTargetId(String targetLabel) {
...
   private List<AnalysisProtosV2.Artifact> expandDepsetToArtifacts(Integer idToExpand) {
-    Queue<Integer> idsToExpand =
-        new ArrayDeque<Integer>() {
-          {
-            add(idToExpand);
-          }
-        };
+    Queue<Integer> idsToExpand = new ArrayDeque<>(Lists.newArrayList(idToExpand));
 
     HashSet<Integer> expandedIds = new HashSet<>();
```
gerardd33, line 74: `Set<Integer> ... = `

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
```diff
@@ -64,8 +64,9 @@ private String getTargetId(String targetLabel) {
       expandedIds.add(depsetId);
 
       actionGraph.getDepSetOfFilesList().stream()
-              .filter((depset) -> depsetId.equals(depset.getId()))
-              .forEach((depset) -> {
+          .filter((depset) -> depsetId.equals(depset.getId()))
```
gerardd33, line 67: `depset` instead of `(depset)`. A bit below too.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -82,8 +82,9 @@ private int getTargetId(String targetLabel) {
       expandedIds.add(depsetId);
 
       actionGraph.getDepSetOfFilesList().stream()
-              .filter((depset) -> depsetId.equals(depset.getId()))
-              .forEach((depset) -> {
+          .filter((depset) -> depsetId.equals(depset.getId()))
```
gerardd33, line 85: Same as above.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
gerardd33: I don't think it's that simple. Later there is this while that does `idsToExpand.addAll(` to this queue.

gerardd33: It looks like some kind of BFS.

Abrams27: ohh right, sorry

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -84,13 +80,13 @@ private int getTargetId(String targetLabel) {
         return Collections.emptyList();
```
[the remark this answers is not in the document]
andrefmrocha: The return here is wrong as a whole, since that breaks the bfs. It doesn't seem to be common for duplicated to show up, but it's best to be safe.

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
agluszak: I should have expressed it in a better way, but what I mean is not the specification of the proto itself, but rather explanation of what this parser does. Currently it appears to be a black box. As you mentioned in one of your comments, this parser shares a lot of logic with the V1 parser, but it is not clear what this logic is supposed to do.

agluszak: Regarding the unit tests, what I meant was related to the problem with the queue. There certainly was a mistake within the method logic (new queue with a single element being created each time), but we didn't have the tests which would allow us to spot this mistake. The tests could also act as a specification. I would like to know what's the expected output of all these methods - and I could potentially learn that by reading the tests, but there are none

agluszak: I'm not sure what you mean by example binaries though. I imagine that these tests could use some hand-crafted instances of the class `AnalysisProtosV2.ActionGraphContainer`

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,96 @@
...
+  @Override
+  protected Stream<String> buildInputs(String target, List<String> suffixes) {
+    return getActions(target).stream()
+        .flatMap(action -> action.getInputDepSetIdsList().stream())
+        .flatMap(depset -> expandDepsetToArtifacts(depset).stream())
+        .map(AnalysisProtosV2.Artifact::getPathFragmentId)
+        .map(pathFragmentId -> EXEC_ROOT + constructPath(pathFragmentId));
+  }
+
+  private String constructPath(Integer pathFragmentId) {
```
agluszak, line 29: I would love to see a more informative name for this method. What kind of path it constructs?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public ActionGraphV2Parser(AnalysisProtosV2.ActionGraphContainer actionGraph) {
+    this.actionGraph = actionGraph;
+    pathFragmentMap =
+        actionGraph.getPathFragmentsList().stream()
+            .collect(Collectors.toMap(AnalysisProtosV2.PathFragment::getId, fragment -> fragment));
+  }
+
+  @Override
+  protected Stream<String> buildInputs(String target, List<String> suffixes) {
```
agluszak, line 21: what kind of inputs are built by this method?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,96 @@
...
+      pathBuilder.add(fragment.getLabel());
+      currId = fragment.getParentId();
+    }
+
+    Collections.reverse(pathBuilder);
+    return String.join("/", pathBuilder);
+  }
+
+  @Override
+  public List<String> getOutputs(String target, List<String> suffixes) {
```
agluszak, line 43: same as above

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
```diff
@@ -0,0 +1,96 @@
...
+            .collect(Collectors.toSet());
+    return actionGraph.getArtifactsList().stream()
+        .filter(artifact -> artifactIds.contains(artifact.getId()))
+        .map(AnalysisProtosV2.Artifact::getPathFragmentId)
+        .map(this::constructPath)
+        .filter(path -> suffixes.stream().anyMatch(path::endsWith))
+        .collect(Collectors.toList());
+  }
+
+  private int getTargetId(String targetLabel) {
```
agluszak, line 56: wouldn't `findTargetIdForLabel` be a better name?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
agluszak: could you add a comment explaining what this method does?

agluszak (changes requested):

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: I'll leave some comments on the abstract class to guarantee that the outputs and inputs of a action graph parser are clear

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: same as above

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: I've left a comment to better explain this

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: I'm not sure if I can a better one as of right now... This function is quite generic in the sense that it builds a path from a path fragment, which is a way of an artifact expressing a portion of its path to the file it represents. Can you think of a way of naming this better?

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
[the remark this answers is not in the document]
agluszak: Done. https://github.com/JetBrains/bazel-bsp/pull/45

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
agluszak: yup, please do

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
agluszak: only one should be removed imo

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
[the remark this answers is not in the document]
andrefmrocha: Cool! Going to resolve this then

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV2Parser.java
andrefmrocha: It has already been changed to `targetLabel`

src/test/java/org/jetbrains/bsp/bazel/BazelBspServerTest.java
andrefmrocha: Since the creation of the test clients changed to the `run` method, I think this sort of logging should be done there. I've changed it up a bit. Let me know what you think

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
```diff
@@ -1,45 +1,28 @@
...
+package org.jetbrains.bsp.bazel.server.bsp.resolvers.actiongraph;
 
 import com.google.common.collect.Lists;
 import com.google.devtools.build.lib.analysis.AnalysisProtos;
-import java.util.ArrayDeque;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Queue;
-import java.util.Set;
+import java.util.*;
```
gerardd33, line 5: Where will this be resolved? What did you guys decide?

gerardd33 (approved):
Looks great

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
andrefmrocha: #45 is going to fix that

Abrams27 (approved):
lets go, great job!

src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/actiongraph/ActionGraphV1Parser.java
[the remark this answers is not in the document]
andrefmrocha: I've went ahead and merged dev, now that #42 has been fixed.

PR #45, opened 2021-03-14
Prohibit wildcard imports using error-prone

agluszak:
So apparently error-prone is built into Bazel and provides a check for wildcard imports.
http://errorprone.info/docs/installation

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,2 +1,5 @@
 build --java_toolchain=@bazel_tools//tools/jdk:toolchain_java8
 build --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java8
+build --javacopt="-Xep:WildcardImport:ERROR"
+test --javacopt="-Xep:WildcardImport:ERROR"
+run --javacopt="-Xep:WildcardImport:ERROR"


andrefmrocha:
I don't think BSP will handle this sort of inspections, is there any plugin to do so?

Abrams27 (commented):
fine for me if you tested it (including our implementation)

agluszak:
> I don't think BSP will handle this sort of inspections, is there any plugin to do so?
what do you mean @andrefmrocha? why would BSP have to handle this inspection?
if you're asking how this works - with this flag enabled error-prone check turns using wildcard imports into javac errors and they are reported just as any other javac error would be

Abrams27:
it looks like this:

image

andrefmrocha:
> (btw dont know why this is success)

just because the code doesn't compile, it doesn't mean the import should return as a failure

Abrams27:
> > (btw dont know why this is success)
> 
> just because the code doesn't compile, it doesn't mean the import should return as a failure

right

agluszak:
I still don't understand where the problem is. If you want to use error-prone with Intellij, you can use this plugin https://plugins.jetbrains.com/plugin/7349-error-prone-compiler
This PR is not related to BSP itself in any way. Its purpose is to make the CI fail when  wildcard imports are detected in code

agluszak:
> just because the code doesn't compile, it doesn't mean the import should return as a failure

Ahh, okay, now I understand what you mean. Well, this is a problem with our implementation of the bsp server ;) because it runs compile during import it will fail if there are *any* compilation errors. This is not specific to error-prone. As I said earlier, error-prone plugs itself into the java compiler and emits warnings/errors the same way javac would

andrefmrocha (approved):

PR #49, opened 2021-03-23
Fix build calls and outgroup groups interpretation

andrefmrocha:
Fixes currently existing problems with our CI.

The problem was found to be due to some changes made to how multiple targets are handled by the `build` command. 

The build command does not accept mnemonics and multiple targets should be treated as multiple arguments.

Furthermore, there was a constant wrong in the `ClasspathParser`


Abrams27:
Well, to be honest I don't get the idea, there's why:
- we use `withMnemonic()` only once with `aquery()`, there is no `build().withMnemonic()` call chain
- probably there are possible other similar wrong `<command>().with*()` calls, but to address all of them, we would have to create separate builders for every command - tbh it's a great idea and probably needed in the longer run, but now it's imo waste of time
-  even if we want to do something similar for the `build` command, it should be done in another way - overriding `withTargets` to `withArguments` makes no sense and introduces ambiguity

andrefmrocha:
> * we use `withMnemonic()` only once with `aquery()`, there is no `build().withMnemonic()` call chain

The gathering of targets with the `+` is also a type of mnemonic, so the problem is in the name `withMnemonic` name on the builder.

> * probably there are possible other similar wrong `<command>().with*()` calls, but to address all of them, we would have to create separate builders for every command - tbh it's a great idea and probably needed in the longer run, but now it's imo waste of time
> * even if we want to do something similar for the `build` command, it should be done in another way - overriding `withTargets` to `withArguments` makes no sense and introduces ambiguity

I agree with you, I just wanted to see how these made out in the CI by taking the easy way out, I did not mark it as a draft and that's my bad.

Abrams27:
> > * we use `withMnemonic()` only once with `aquery()`, there is no `build().withMnemonic()` call chain
> 
> The gathering of targets with the `+` is also a type of mnemonic, so the problem is in the name `withMnemonic` name on the builder.
> 
> > * probably there are possible other similar wrong `<command>().with*()` calls, but to address all of them, we would have to create separate builders for every command - tbh it's a great idea and probably needed in the longer run, but now it's imo waste of time
> > * even if we want to do something similar for the `build` command, it should be done in another way - overriding `withTargets` to `withArguments` makes no sense and introduces ambiguity
> 
> I agree with you, I just wanted to see how these made out in the CI by taking the easy way out, I did not mark it as a draft and that's my bad.

okeyyy, unfortunately CI failed again : (

andrefmrocha:
> okeyyy, unfortunately CI failed again : (

Yap, I'm looking into it more deeply tomorrow. Hopefully before our meeting we'll have a fix!

Abrams27:
> > > * we use `withMnemonic()` only once with `aquery()`, there is no `build().withMnemonic()` call chain
> > 
> > 
> > The gathering of targets with the `+` is also a type of mnemonic, so the problem is in the name `withMnemonic` name on the builder.

+ it (hopefully) mapped 1:1 old bazel calls, and CI was passing after introducing bazel runner. 

Btw, last week I and Andrzej did a little investigation and we finished at the `getScalaBuildTarget` method and `buildTargetsWithBep` call, unfortunately we had to finish then. Maybe it could be good starting point for the further investigation

andrefmrocha:
> Btw, last week I and Andrzej did a little investigation and we finished at the `getScalaBuildTarget` method and `buildTargetsWithBep` call, unfortunately we had to finish then. Maybe it's a good starting point for the further investigation

This was the basis of this change because there was this log whenever a scala bazel project was opened:
```
Command aborted with reason LOADING_FAILURE: The package part of '(@io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library+@io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect+@io_bazel_rules_scala_scala_compiler//:io_bazel_rules_scala_scala_compiler)' should not end in a slash
```

After executing the command of the aspect manually:
```
bazel build --aspect=<aspect_name> (@io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library+@io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect+@io_bazel_rules_scala_scala_compiler//:io_bazel_rules_scala_scala_compiler)
```
I got the same error. However, when I changed to:
```
bazel build --aspect=<aspect_name> @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library @io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect@io_bazel_rules_scala_scala_compiler//:io_bazel_rules_scala_scala_compiler
```

The command worked perfectly fine. After a quick check in `master`, I was able to find that we did not group targets with a `+` there when it came to building, only in queries.

Abrams27:
> > Btw, last week I and Andrzej did a little investigation and we finished at the `getScalaBuildTarget` method and `buildTargetsWithBep` call, unfortunately we had to finish then. Maybe it's a good starting point for the further investigation
> 
> This was the basis of this change because there was this log whenever a scala bazel project was opened:
> 
> ```
> Command aborted with reason LOADING_FAILURE: The package part of '(@io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library+@io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect+@io_bazel_rules_scala_scala_compiler//:io_bazel_rules_scala_scala_compiler)' should not end in a slash
> ```
> 
> After executing the command of the aspect manually:
> 
> ```
> bazel build --aspect=<aspect_name> (@io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library+@io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect+@io_bazel_rules_scala_scala_compiler//:io_bazel_rules_scala_scala_compiler)
> ```
> 
> I got the same error. However, when I changed to:
> 
> ```
> bazel build --aspect=<aspect_name> @io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library @io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect@io_bazel_rules_scala_scala_compiler//:io_bazel_rules_scala_scala_compiler
> ```
> 
> The command worked perfectly fine. After a quick check in `master`, I was able to find that we did not group targets with a `+` there when it came to building, only in queries.

ohhh sounds great, maybe easier solution would be changing `withTargets` into `withArguments` at the `buildTargetsWithBep` method

andrefmrocha:
> ohhh sounds great, maybe easier solution would be changing `withTargets` into `withArguments` at the `buildTargetsWithBep` method

I think that can be a quick fix as well. That's more passabale downstream, but we probably need to look into the semantics of the builder as well.

andrefmrocha:
This, in fact, fixes the problem in question. But there's something wrong with the BEP output group. Going to investigate that

andrefmrocha:
> This, in fact, fixes the problem in question. But there's something wrong with the BEP output group. Going to investigate that

This turned out to be the source of the problem. I've applied @Abrams27 suggestion as well. Do you guys mind having a quick look into this? With this, we can also merge #42 and I can finish #43 and finish the cpp extension.

Abrams27 (approved):

PR #51, opened 2021-03-29
>=info logs are separate now

Abrams27:
I decided (just for the clarity) separate logs with level >=info to another file and creating file `.all.log` for all logs. It would be useful for checking entire flow without unnecessary events etc.


andrefmrocha:
Are we planning on adding more logs as debug logs in the future?

Abrams27:
> Are we planning on adding more logs as debug logs in the future?

I hope so, but at this moment it's just for flow clarity - bep events make following import flow a bit difficult

andrefmrocha (approved):

PR #53, opened 2021-04-16
Entire repository import test

Abrams27:
Hopefully it does test something


test-resources/.bazelbsp/aspects.bzl
```diff
@@ -0,0 +1,40 @@
+def _print_aspect_impl(target, ctx):
```
andrefmrocha, line 1: Isn't this entire folder not supposed to be here?

test-resources/.bsp/bazelbsp.json
```diff
@@ -0,0 +1,17 @@
+{
```
andrefmrocha, line 1: Same with this folder

andrefmrocha (changes requested):
A concern was raised

andrefmrocha (approved):
lgmt!

PR #56, opened 2021-05-27
Guarantee deps are built before finishing import.

andrefmrocha:
Dependencies must be fully built in order to guarantee the dependencies are all present for indexing.

Opening this one as a PR since I'm not sure if there's a better way of doing this.

### DIFF
--- a/src/main/java/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspCompilationManager.java
+++ b/src/main/java/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspCompilationManager.java
@@ -49,16 +49,19 @@ public class BazelBspCompilationManager {
 
     Build.QueryResult queryResult = QueryResolver.getQueryResultForProcess(bazelProcess);
 
-    for (Build.Target target : queryResult.getTargetList()) {
-      target.getRule().getRuleOutputList().stream()
-          .filter(output -> output.contains(Constants.DIAGNOSTICS))
-          .forEach(
-              output ->
-                  diagnosticsProtosLocations.put(
-                      target.getRule().getName(),
-                      BuildManagerParsingUtils.convertOutputToPath(
-                          output, bazelData.getBinRoot())));
-    }
+    queryResult.getTargetList().stream()
+        .map(Build.Target::getRule)
+        .filter(rule -> !rule.getName().startsWith("@"))
+        .forEach(
+            rule ->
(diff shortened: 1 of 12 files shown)


src/main/java/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspQueryManager.java
```diff
@@ -146,6 +146,33 @@ private BuildTarget getBuildTargetForRule(Build.Rule rule) {
         .collect(Collectors.toList());
   }
 
+  public List<BuildTargetIdentifier> getTargetIdentifiersForDependencies(
+      List<BuildTargetIdentifier> targets) {
+    String targetsUnion =
```
Abrams27, line 151: It looks like a good thing to move to the `TargetsUtils`

src/main/java/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspCompilationManager.java
```diff
@@ -49,16 +49,19 @@ public BazelBspCompilationManager(BazelRunner bazelRunner, BazelData bazelData)
...
-      target.getRule().getRuleOutputList().stream()
-          .filter(output -> output.contains(Constants.DIAGNOSTICS))
-          .forEach(
-              output ->
-                  diagnosticsProtosLocations.put(
-                      target.getRule().getName(),
-                      BuildManagerParsingUtils.convertOutputToPath(
-                          output, bazelData.getBinRoot())));
-    }
+    queryResult.getTargetList().stream()
```
Abrams27, line 52: this sequence becomes long and complex, maybe it should be splitted somehow

Abrams27 (commented):
hopefully it solves the problem, left minor comments

src/main/java/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspQueryManager.java
andrefmrocha: the whole method or just getting the union? I'm not sure what you mean from your selection  :sweat_smile:

Abrams27: this part: `targets.stream().map(BuildTargetIdentifier::getUri).collect(Collectors.joining("+"));`

PR #123, opened 2021-12-21
Allow running a Bazel command without [positional] arguments

### LINKED ISSUE #119: Rebuilding project doesn't work
This is what I see when I try to rebuild project:

image

odisseus:
One command that doesn't accept arguments at all (only flags) is `bazel clean`.

Fixes #119.


abrams27 (approved):
solves the issue, nevertheless bazel runner needs to be updated in the future anyway (https://github.com/JetBrains/bazel-bsp/issues/124)

PR #126, opened 2021-12-28
Better heuristics for guessing source roots

### LINKED ISSUE #115: Source roots approximation is incorrect
I had a root directory in my workspace called java_hello. `WORKSPACE_ROOT/java_hello/src/main/java/com/hello/Hello.java`. 
The heuristics for detecting source roots found `java` in `java_hello` and inferred root to be `WORKSPACE_ROOT/java` - a directory which doesn't exist. Which resolved in package prefix `..java_hello.src.main.java.com.hello` instead of expected `com.hello`.
In fastpass we've been using following logic, which seemed to work fine:

```scala
  private val sourceRootPattern = FileSystems.getDefault.getPathMatcher(
    "glob:**/{main,test,tests,src,3rdparty,3rd_party,thirdparty,third_party}/{resources,scala,java,jvm,proto,python,protobuf,py}"
  )
  private val defaultTestRootPattern = FileSystems.getDefault.getPathMatcher(
    "glob:**/{test,tests}"
  )

  private def approximateSourceRoot(dir: Path): Option[Path] = {
    @tailrec def loop(d: Path): Option[Path] = {
      if (sourceRootPattern.matches(d)) Some(d)
      else if (defaultTestRootPattern.matches(d)) Some(d)
      else {
        Option(d.getParent) match {
          case Some(parent) => loop(parent)
          case None => None
        }
      }
    }
    loop(dir)
  }
```

Maybe it might be configurable. But actually it would be the best if bazel could provide this information.

odisseus:
Fixes #115.

Credits for the idea go to @lukaszwawrzyk.

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
@@ -6,9 +6,7 @@ import ch.epfl.scala.bsp4j.SourceItem;
 import ch.epfl.scala.bsp4j.WorkspaceBuildTargetsResult;
 import com.google.common.collect.ImmutableList;
 import com.google.devtools.build.lib.query2.proto.proto2api.Build;
-
 import java.net.URI;
-import java.net.URL;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.Path;
@@ -18,11 +16,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
-
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
 import org.eclipse.lsp4j.jsonrpc.messages.ResponseError;
-import org.jetbrains.bsp.bazel.bazelrunner.BazelRunner;
-import org.jetbrains.bsp.bazel.bazelrunner.data.BazelData;
 import org.jetbrains.bsp.bazel.commons.Constants;
 import org.jetbrains.bsp.bazel.commons.Lazy;
 import org.jetbrains.bsp.bazel.server.bep.BepServer;
@@ -37,7 +32,6 @@ public class BazelBspServerBuildManager {
   public static final String BAZEL_PRINT_ASPECT = "@//.bazelbsp:aspects.bzl%print_aspect";
 
   private final BazelBspServerRequestHelpers serverRequestHelpers;
-  private final BazelData bazelData;
   private final BazelBspQueryManager bazelBspQueryManager;
   private final BazelBspCompilationManager bazelBspCompilationManager;
   private final BazelBspTargetManager bazelBspTargetManager;
@@ -48,15 +42,12 @@ public class BazelBspServerBuildManager {
 
   public BazelBspServerBuildManager(
       BazelBspServerRequestHelpers serverRequestHelpers,
-      BazelData bazelData,
-      BazelRunner bazelRunner,
       BazelBspCompilationManager bazelBspCompilationManager,
       BazelBspAspectsManager bazelBspAspectsManager,
       BazelBspTargetManager bazelBspTargetManager,
       BazelCppTargetManager bazelCppTargetManager,
       BazelBspQueryManager bazelBspQueryManager) {
     this.serverRequestHelpers = serverRequestHelpers;
-    this.bazelData = bazelData;
     this.bazelBspCompilationManager = bazelBspCompilationManager;
     this.bazelBspAspectsManager = bazelBspAspectsManager;
     this.bazelCppTargetManager = bazelCppTargetManager;
@@ -76,17 +67,17 @@ public class BazelBspServerBuildManager {
   public String getSourcesRoot(URI sourceUri) {
     Path sourcePath = Paths.get(sourceUri);
     FileSystem fs = FileSystems.getDefault();
-    PathMatcher sourceRootPattern = fs.getPathMatcher("glob:**/" +
-            "{main,test,tests,src,3rdparty,3rd_party,thirdparty,third_party}/" +
-            "{resources,scala,java,kotlin,jvm,proto,python,protobuf,py}");
+    PathMatcher sourceRootPattern = fs.getPathMatcher(
+            "glob:**/"
+                    + "{main,test,tests,src,3rdparty,3rd_party,thirdparty,third_party}/"
+                    + "{resources,scala,java,kotlin,jvm,proto,python,protobuf,py}");
     PathMatcher defaultTestRootPattern = fs.getPathMatcher("glob:**/{test,tests}");
     Path sourceRootGuess = null;
-    for(PathMatcher pattern : new PathMatcher[]{sourceRootPattern, defaultTestRootPattern}){
+    for (PathMatcher pattern : new PathMatcher[]{sourceRootPattern, defaultTestRootPattern}) {
       sourceRootGuess = approximateSourceRoot(sourcePath, pattern);
-      if(sourceRootGuess != null)
-        break;
+      if (sourceRootGuess != null) break;
     }
-    if(sourceRootGuess == null) {
+    if (sourceRootGuess == null) {
       return sourcePath.getParent().toString();
     }
     return sourceRootGuess.toAbsolutePath().toString();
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
```diff
@@ -161,31 +164,49 @@ public void onBuildExit() {
...
+              } catch (Exception e) {
+                throw new RuntimeException(e);
+              }
+            })
+            .map(serverBuildManager::getSourcesRoot)
+            .map(Uri::fromAbsolutePath)
+            .map(uri -> uri.toString())
+            .collect(Collectors.toSet());
+
+    return ImmutableList.copyOf(sourceRootUris);
```
abrams27, line 209:
```suggestion
  private List<String> getRuleRoots(List<SourceItem> items) {
   return items.stream()
            .map(SourceItem ::getUri)
            .map(uri -> {
              try {
                URL url = new URL(uri);
                return url.toURI();
              } catch (Exception e) {
                throw new RuntimeException(e);
              }
            })
            .map(serverBuildManager::getSourcesRoot)
            .map(Uri::fromAbsolutePath)
            .map(Uri:toString)
            .distinct()
            .collect(Collectors.toList());
```

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
```diff
@@ -64,13 +64,32 @@ public BazelBspServerBuildManager(
...
+    PathMatcher defaultTestRootPattern = fs.getPathMatcher("glob:**/{test,tests}");
+    Path sourceRootGuess = null;
+    for (PathMatcher pattern : new PathMatcher[]{sourceRootPattern, defaultTestRootPattern}) {
+      sourceRootGuess = approximateSourceRoot(sourcePath, pattern);
+      if (sourceRootGuess != null) break;
+    }
+    if (sourceRootGuess == null) {
+      return sourcePath.getParent().toString();
+    }
+    return sourceRootGuess.toAbsolutePath().toString();
```
abrams27, line 83:
```suggestion
Optional<Path> patternPath = ImmutableList.of(sourceRootPattern, defaultTestRootPattern).stream()
      .map(pattern -> approximateSourceRoot(sourcePath, pattern))
      .filter(Optional::isPresent)
      .map(Optional::get)
      .findFirst()
      .map(Path::toAbsolutePath);

  return patternPath.orElse(sourcePath.getParent())
      .toString();
```

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
```diff
@@ -64,13 +64,32 @@ public BazelBspServerBuildManager(
...
+  }
+
+  private Path approximateSourceRoot(Path dir, PathMatcher matcher) {
```
abrams27, line 86: i believe that it would be better to use `Optional` and return `Empty` instead of `null`. Also, it would be more efficient to change recursion into a loop

abrams27 (changes requested):
thanks a lot for the contribution!

i left some comments, also it would be nice to add some tests (I know you don't see a lot of tests now, but I'm trying to change that)

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
odisseus: Reformatted

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
odisseus: I think that my variant is more readable, even though it uses null.

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
odisseus: Removed the recursion. Also, I think it's okay to use null here, because this is just a private method, and only a few lines long. On the other hand, the `getParent` method may return null anyway, and I'd rather not introduce Java optionals just for the sake of it.

odisseus:
I have extracted the source root heuristic into a separate class and added some test cases. What exactly should I write in the changelog?

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
abrams27: i strongly disagree - at least for me it looks very messy

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
```diff
@@ -161,31 +164,51 @@ public void onBuildExit() {
...
+                  try {
+                    URL url = new URL(uri);
+                    return url.toURI();
+                  } catch (Exception e) {
+                    throw new RuntimeException(e);
+                  }
+                })
+            .map(SourceRootGuesser::getSourcesRoot)
+            .map(Uri::fromAbsolutePath)
+            .map(uri -> uri.toString())
```
abrams27, line 208: lambda could be replaced with a method reference

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
```diff
@@ -161,31 +164,51 @@ public void onBuildExit() {
...
 
-  private List<String> getRuleRoots(Build.Rule rule) {
-    String sourcesRootUriString = serverBuildManager.getSourcesRoot(rule.getName());
-    Uri uri = Uri.fromAbsolutePath(sourcesRootUriString);
-
-    return ImmutableList.of(uri.toString());
+  private List<String> getRuleRoots(List<SourceItem> items) {
+    Set<String> sourceRootUris =
+        items.stream()
+            .map(item -> item.getUri())
```
abrams27, line 196: lambda could be replaced with a method reference

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
```diff
@@ -161,31 +164,51 @@ public void onBuildExit() {
...
+                  } catch (Exception e) {
+                    throw new RuntimeException(e);
+                  }
+                })
+            .map(SourceRootGuesser::getSourcesRoot)
+            .map(Uri::fromAbsolutePath)
+            .map(uri -> uri.toString())
+            .collect(Collectors.toSet());
+
+    return ImmutableList.copyOf(sourceRootUris);
```
abrams27, line 211: could be replaced by `.distinct()` (look previous version of review)

abrams27 (changes requested):
one more thing from the previous review

---
> What exactly should I write in the changelog?

added better heuristics for guessing source roots / fixed heuristics for guessing source roots - whatever you prefer

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/BazelBspServerBuildManager.java
jastice: Both are readable to me. I don't think we should make such style matters a big issue in accepting PRs ;)

odisseus: Refactored

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
odisseus: Done

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
odisseus: Done

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/BuildServerService.java
odisseus: Done

abrams27 (approved):
Looks great! Thanks so much for the contribution, I really appreciate it!

PR #130, opened 2022-01-12
Handle multiple projects in one workspace

### LINKED ISSUE #122: Support for multiple projects in one repo
I am looking for a solution to have many projects open in parallel on one bazel workspace/vcs root. The goal is to have a directory for projects, it can be inside or outside repo. For example fastpass does REPO/../bsp-projets/PROJECT or our custom integration with bazel and google's plugin does REPO/bazel-imports/PROJECT.
The PROJECT dir should contain the .bsp, .bazelbsp, .idea directories and anything else that is necessary like projectview file or symlinks. IntelliJ would open project in this dir and should still show the whole repo once the bsp project sync completes.
Currently this scenario seems impossible, as I see no path to workspace anywhere, It seems like it is assumed to be .bsp/.. dir. Also I can only place projectview in the root of repository.

lukaszwawrzyk:
Fixes #122.

The goal of this PR is to allow creating the bsp/intellij project root in different location than the repository root.
Example locations could be <REPO>/bsp-projects/<PROJ> or <REPO>/../bsp-projects/<PROJ>.
I imagine it would be used through a wrapper like `bazel idea --bsp --name myproject --targets //target_a/... //target_b/...`, and such wrapper would manage where this directory is, if project already exists, auto generating name and similar things. But first, the server needs to support it.

I made a distinction between `workspace root` (bazel workspace root) and `bsp project root` (root directory for .bsp, .idea, .bazelbsp, projectview.bazelview files). I added workspace root as startup arg of the server and it is set up by the installer. It is used as a directory where bazel is run from. Without this it would work fine if bsp-projects was inside repo root, but if it is outside, it is necessary.

Now by default projectview file is expected to be in the bsp project root.

To use this feature installer should be run from workspace directory as `installer -d bsp-projects/project-name` or `installer -d ~/bsp-projects/project-name -w ~/workspaces/repo` from any directory.

The issue in IntelliJ is that when the workspace and project are in different directories, project tree is not a tree. Instead, each module (except modules inside modules) becomes a root. This lacks context. Ideally we would see the whole repo, or a common parent and all its children that are modules or has modules as children. I made a workaround on `root-module` branch, that creates a fake root module that points to the workspace. I am not opening PR for that because it triggers indexing of whole monorepo and most of files contain highlighting errors. This is thing to resolve later.

This PR depends on Java 11 migration, so I will rebase it after that one is merged.

### DIFF
--- a/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelRunner.java
+++ b/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelRunner.java
@@ -1,9 +1,13 @@
 package org.jetbrains.bsp.bazel.bazelrunner;
 
+import ch.epfl.scala.bsp4j.StatusCode;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.Lists;
+import java.io.File;
 import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Supplier;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jetbrains.bsp.bazel.server.loggers.BuildClientLogger;
@@ -19,9 +23,27 @@ public class BazelRunner {
 
   private Optional<Integer> besBackendPort = Optional.empty();
   private Optional<BuildClientLogger> buildClientLogger = Optional.empty();
+  private final Supplier<File> workspaceRoot;
 
   public BazelRunner(String bazelBinaryPath) {
     this.bazel = bazelBinaryPath;
+    this.workspaceRoot = Suppliers.memoize(this::resolveWorkspaceRoot);
+  }
+
+  private File resolveWorkspaceRoot() {
+    var builder = new ProcessBuilder(bazel, "info", "workspace");
+    try {
+      var process = new BazelProcess(builder.start(), Optional.empty());
+      var result = process.waitAndGetResult();
+      if (result.getStatusCode() != StatusCode.OK) {
+        throw new RuntimeException(
+            "Failed to run bazel info workspace. Make sure that the project is created inside a"
+                + " bazel workspace");
+      }
+      return new File(result.getStdout().get(0));
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
   }
 
   public BazelRunnerCommandBuilder commandBuilder() {
@@ -64,6 +86,7 @@ public class BazelRunner {
     LOGGER.info("Running: {}", processArgs);
 
     ProcessBuilder processBuilder = new ProcessBuilder(processArgs);
+    processBuilder.directory(workspaceRoot.get());
     Process process = processBuilder.start();
 
     return new BazelProcess(process, buildClientLogger);
--- a/install/src/main/java/org/jetbrains/bsp/bazel/install/Install.java
+++ b/install/src/main/java/org/jetbrains/bsp/bazel/install/Install.java
@@ -64,8 +64,7 @@ public class Install {
         Path rootDir = getRootDir(cmd);
         writeConfigurationFiles(rootDir, details);
 
-        String currentDirectory = getCurrentDirectory();
-        System.out.println("Bazel BSP server installed in '" + currentDirectory + "'.");
+        System.out.println("Bazel BSP server installed in '" + rootDir.toAbsolutePath() + "'.");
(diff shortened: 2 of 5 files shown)


abrams27:
Thanks for it! Nevertheless, I think that this solution is a little over the top.

Okey, soo - I think that it could be done way more easier.
1) I would like to move all the configuration to project view - including current installer flags (I'm working on it right now). Then installer would take only path to the project view file as an argument. (It'll be useful - project view could be read at initialization and workspace reload). In project view you targets and directories could easily specified.
2) BSP specification allows to create more than one bsp configuration file in the `.bsp` directory. Only names have to unique.
3) We can create multiple server instances with multiple configuration files. Each server installer will take path to a project view file, so we can use it in name - e.g. `bazel-bsp-path/to/projectview.bazelproject`

Finally, we can connect to each server, and work in parallel on one bazel workspace.

I think that this solution is more elegant and much easier to implement (minor changes in the installer)
Does your solution solve any problem that I missed?

--- 
> The issue in IntelliJ is that when the workspace and project are in different directories, project tree is not a tree. Instead, each module (except modules inside modules) becomes a root. This lacks context. Ideally we would see the whole repo, or a common parent and all its children that are modules or has modules as children. I made a workaround on `root-module` branch, that creates a fake root module that points to the workspace. I am not opening PR for that because it triggers indexing of whole monorepo and most of files contain highlighting errors. This is thing to resolve later.

I'm afraid that it is caused by the IntelliJ client itself.

abrams27:
(Obviously this mechanism will be useful anyway)

install/src/main/java/org/jetbrains/bsp/bazel/install/Install.java
```diff
@@ -83,10 +86,28 @@ public static void main(String[] args) throws IOException {
...
-        : Paths.get("");
+        ? Paths.get(cmd.getOptionValue(DIRECTORY_SHORT_OPT)).toAbsolutePath().normalize()
+        : workspace;
+  }
+
+  private static Path resolveBazelWorkspace() {
+    return resolveBazelWorkspace(Paths.get("").toAbsolutePath());
+  }
+
+  private static Path resolveBazelWorkspace(Path dir) {
```
abrams27, line 99: I believe that it could be done using bazel - `bazel info workspace`

abrams27 (commented):
I think that you can now rebase do java bump

---
Btw, this change and my workspace model change (with backward compatibility) I would like to release as `1.1.1` and then start working on the multi project mechanism as `2.0.0`

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelRunner.java
```diff
@@ -19,9 +23,27 @@
...
+  private File resolveWorkspaceRoot() {
+    var builder = new ProcessBuilder(bazel, "info", "workspace");
+    try {
+      var process = new BazelProcess(builder.start(), Optional.empty());
+      var result = process.waitAndGetResult();
+      if (result.getStatusCode() != StatusCode.OK) {
+        throw new RuntimeException(
+            "Failed to run bazel info workspace. Make sure that the project is created inside a"
+                + " bazel workspace");
+      }
```
abrams27, line 42:
```
bazelRunner
    .commandBuilder()
    .info()
    .withArgument(argument)
    .executeBazelCommand()
    .waitAndGetResult();
```

abrams27 (changes requested):
we have a bazel builder for bazel calls, rest looks great

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelRunner.java
lukaszwawrzyk: But this is the bazel runner that I am changing :p I want to make sure that bazel is run from workspace root to avoid problems with potential relative paths to targets. To achieve this, bazel runner must create process with cwd set to workspace root. I need to first resolve what this root is. I can't use bazel runner for it as it is not yet ready because it doesn't have this information.

abrams27: ohh right, then it sucks a bit (i mean bazel runner, not your fault), okey so we can merge it

abrams27 (approved):

PR #131, opened 2022-01-17
Filter out non runtime jars for running apps and tests

lukaszwawrzyk:
TargetsLanguageOptionsResolver that I am using returns all outputs and inputs for given target. This includes source jars that are useless at runtime as well as hjars and ijars. For example this causes junit classes without bodies to appear on classpath and shadowing the actual implementation which fails tests at runtime. 
I proposed a simple workaround here, to just remove jars that *look like* not relevant for runtime. This fixed test running for me.
I feel like it is not *the solution*, I would like to know from bazel if a jar is relevant for runtime or not, but I don't know how to get  this.
I see that we might try the custom aspect approach from google's plugin: https://github.com/bazelbuild/intellij/blob/master/aspect/java_classpath.bzl but I hoped it should be easier than custom aspects during building...

### DIFF
--- a/e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestBaseScenario.java
+++ b/e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestBaseScenario.java
@@ -60,7 +60,11 @@ public abstract class BazelBspTestBaseScenario {
   }
 
   private boolean executeScenarioSteps() {
-    return getScenarioSteps().stream().allMatch(BazelBspTestScenarioStep::executeAndReturnResult);
+    return getScenarioSteps().stream()
+        .map(BazelBspTestScenarioStep::executeAndReturnResult)
+        .collect(java.util.stream.Collectors.toList())
+        .stream()
+        .allMatch(x -> x);
   }
 
   protected abstract List<BazelBspTestScenarioStep> getScenarioSteps();
--- a/e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestScenarioStep.java
+++ b/e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestScenarioStep.java
@@ -22,7 +22,7 @@ public class BazelBspTestScenarioStep {
 
     return Try.run(testkitCall)
         .onSuccess(e -> LOGGER.info("Step \"{}\" executed correctly!", testName))
-        .onFailure(e -> LOGGER.error("Step \"{}\" execution failed! Exception: {}", testName, e))
+        .onFailure(e -> LOGGER.error("Step \"{}\" execution failed!", testName, e))
         .map(i -> true)
         .getOrElse(false);
   }
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/JvmBuildServerService.java
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/JvmBuildServerService.java
@@ -24,6 +24,8 @@ public class JvmBuildServerService {
   private static final Logger LOGGER = LogManager.getLogger(JvmBuildServerService.class);
   private static final ImmutableList<String> JVM_LANGUAGES_IDS =
       ImmutableList.of(Constants.JAVAC, Constants.KOTLINC, Constants.SCALAC);
+  private static final ImmutableList<String> NON_RUNTIME_SUFFIXES =
+      ImmutableList.of("-hjar.jar", "-sources.jar", "-src.jar", "-ijar.jar");
 
   private final TargetsLanguageOptionsResolver<JvmEnvironmentItem> targetsLanguageOptionsResolver;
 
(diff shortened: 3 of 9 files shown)


lukaszwawrzyk:
I managed to do this with aspects. I also noticed that aspect don't work if bsp workspace root != bazel workspace root. I will fix it later.

brown:
There's also a Java language server that can extract information from Bazel:  https://github.com/georgewfraser/java-language-server

I believe is also uses aspects.

e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestBaseScenario.java
```diff
@@ -60,7 +60,11 @@ public void executeScenario() {
   }
 
   private boolean executeScenarioSteps() {
-    return getScenarioSteps().stream().allMatch(BazelBspTestScenarioStep::executeAndReturnResult);
+    return getScenarioSteps().stream()
```
abrams27, line 63: what does it do?

e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestScenarioStep.java
```diff
@@ -22,7 +22,7 @@ public boolean executeAndReturnResult() {
 
     return Try.run(testkitCall)
         .onSuccess(e -> LOGGER.info("Step \"{}\" executed correctly!", testName))
-        .onFailure(e -> LOGGER.error("Step \"{}\" execution failed! Exception: {}", testName, e))
+        .onFailure(e -> LOGGER.error("Step \"" + testName + "\" execution failed!", e))
```
abrams27, line 25: why?

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/JvmBuildServerService.java
```diff
@@ -58,8 +69,18 @@ public JvmBuildServerService(BazelData bazelData, BazelRunner bazelRunner) {
...
   private List<JvmEnvironmentItem> getJvmEnvironmentItems(List<BuildTargetIdentifier> targets) {
-    return targetsLanguageOptionsResolver.getResultItemsForTargets(targets).stream()
-        .distinct()
-        .collect(Collectors.toList());
+    var items =
+        targetsLanguageOptionsResolver.getResultItemsForTargets(targets).stream()
+            .distinct()
+            .collect(Collectors.toList());
+    items.forEach(item -> item.setClasspath(fetchClasspath(item.getTarget())));
+    return items;
```
abrams27, line 77:
```suggestion
    return targetsLanguageOptionsResolver.getResultItemsForTargets(targets).stream()
            .distinct()
            .peek(item -> item.setClasspath(fetchClasspath(item.getTarget())))
            .collect(Collectors.toList());
```

abrams27 (commented):
I think that we need to use aspects more, great changes!

(btw, now I see how messy the implementation is)

please rebase

e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestScenarioStep.java
lukaszwawrzyk: To use the overload that accepts `Exception` so that it can be printed properly with stacktrace. Currently it is just `toString()`.

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/JvmBuildServerService.java
lukaszwawrzyk: I did that originally, but this is actually modifying the objects in the collection at some unpredicted time. This means that it affects the calculation of disticts and yields incorrect results.

abrams27: right, stream things

e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestScenarioStep.java
abrams27: ohh, yea, but first placeholder still can be used
```
.onFailure(e -> LOGGER.error("Step \"{}\" execution failed!", testName, e))
```

e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestBaseScenario.java
abrams27: ohh I see, sounds okey (maybe `.parallelStream()` might be a solution as well?)

lukaszwawrzyk: I think that scenario `steps` implies it is done serially and I think some tests may rely in this property. In general it could be unsafe to run some further step if previous failed, but if they turn out to be independent it is better to know that they are ok. Also not sure if they will be any faster in parallel as bazel seems to tell other invocations to wait for the current one to finish.

e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestScenarioStep.java
lukaszwawrzyk: I didn't suspect someone will check the type of last argument, but it seems to be the case, so I will fix this :p

PR #132, opened 2022-01-19
Handle aspects when multiple projects are in workspace

lukaszwawrzyk:
Path to aspects was absolute and hardcoded in multiple places as //.bazelbsp:aspects.bzl%ASPECT_NAME and this only works if workspace root == bsp root. I fixed this and made some refactoring.


abrams27 (approved):
wow, really nice work, thanks!

please rebase and fix formatting

abrams27 (changes requested):
aspects in `server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/services/JvmBuildServerService.java` need change as well

abrams27 (approved):

PR #133, opened 2022-01-20
Add root module to aggregate all modules

lukaszwawrzyk:
When bsp project root is not workspace root, IntelliJ will show all modules on the same level which is not very readable.
This workaround adds the bazel workspace root as a fake module with a single resource root equal to workspace root. Using resources instead of sources is the key to make it work correctly.

Improvement idea for later:
An improvement over this approach is to create a module for each imported target root based on project view. So the top common directory (but not workspace root, unless everything is imported. For example:

```
targets:
  - //a/src/x
  - //a/src/y
  - //b/t
  - //b/q
```

would result in inferring following roots: `a/src` and `b`. For these modules server should respond with a_src-project-root and b_project-root fake modules. This will help to group root modules in the UI and at the same time not show the whole workspace.


abrams27:
I feel like it is a IntelliJ problem- how the client resolves provided targets, (also, creating fake module might also break other clients) how about configuration files from project root (like `WORKSPACE`) in this scenario?

Also, I tried this project view:
```
targets:
  //projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/model
  //projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser
  //bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/data
  //bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/params
```
with the current `dev` branch and 2022.1 IntelliJ and the project structure was still the same, with normal project tree (with "Project" view) 

btw that probably should be changed - we don't want to see not imported directories

lukaszwawrzyk:
I agree it should show just imported roots. This will be much easier with rewrite of project sync that I am doing now.

PR #136, opened 2022-01-24
Return only the relevant output jar rather than 5-6 jars 

lukaszwawrzyk:
When bsp server asks for compile classpath and ouput (javac/scalac options), the server returns a list of identical responses except for class directory. There is one object per output. Bazel reports like 6 outputs jarname.jar jarname-src.jar jarname-native-headers.jar jarname-hjar.jar etc.
On the client side just one response per target is expected. The whole response is converted to map from target id to data which results basically in selecting random/last output from response. 
According to how bazel names these jars and sorting of strings works, my PR will select just the jarname.jar and return single response for a target.

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsLanguageOptionsResolver.java
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsLanguageOptionsResolver.java
@@ -3,10 +3,11 @@ package org.jetbrains.bsp.bazel.server.bsp.resolvers;
 import ch.epfl.scala.bsp4j.BuildTargetIdentifier;
 import com.google.common.collect.ImmutableList;
 import com.google.devtools.build.lib.query2.proto.proto2api.Build;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsLanguageOptionsResolver.java
```diff
@@ -97,6 +98,8 @@ private BazelProcess queryBazel(List<String> targets) {
     List<String> inputs = actionGraphParser.getInputsAsUri(target, bazelData.getExecRoot());
 
     return actionGraphParser.getOutputs(target, ACTION_GRAPH_SUFFIXES).stream()
+        .max(Comparator.naturalOrder())
+        .stream()
```
abrams27, line 102: to reflect that we are returning only one element now it would be nice to leave Optional here, and add `.stream()` it the final `flatMap` in the `getResultItemsForTargets` method

abrams27 (commented):
please rebase to fix CI

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/resolvers/TargetsLanguageOptionsResolver.java
lukaszwawrzyk: right, fixed

Issue #77, opened 2021-08-05
Old Installer command line arguments should be moved to the ProjectView


abrams27:
Recently we introduced ProjectView, which currently allows specifying project's targets and directories. Installer arguments does the same, but also allows changing path to java and bazel and it would be nice to move this configuration to the ProjectView as well. 

`Installer.java` also needs some love`

---
things to do:
- [x] update Project View -- deprecated installer flags | https://github.com/JetBrains/bazel-bsp/pull/143
- [x] `1.1.1` release with deprecated flags 
- [ ] update `Installer.java` -- the only argument should be path to project view, **breaking change, so version `2.0.0`**

PR #143, opened 2022-02-03
[feature] project view 2000

abrams27:
**wowowow project view 2000!!!!**

Introduces completely new project view parsing mechanism (even better one), in the following PR (and releases) usage will be improved (new installer, better initializer and similar).

---

1/n in #77 

---
testing:
```
bazel test //projectview/...
bazel test //commons/...
```

and manually by installing the server with different files


commons/src/main/java/org/jetbrains/bsp/bazel/commons/BUILD
```diff
@@ -7,5 +7,6 @@ java_library(
     deps = [
         "@maven//:ch_epfl_scala_bsp4j",
         "@maven//:com_google_guava_guava",
+        "@maven//:io_vavr_vavr",
```
izzyreal, line 10:
```suggestion
        "@maven//:io_vavr_vavr"
```

commons/src/main/java/org/jetbrains/bsp/bazel/commons/FileUtils.java
```diff
@@ -0,0 +1,18 @@
...
+import java.nio.file.Path;
+
+public final class FileUtils {
+
+  public static String readFileContentOrEmpty(Path filePath) {
+    return Try.success(filePath)
+        .map(Path::toFile)
+        .map(file -> Files.asCharSource(file, Charset.defaultCharset()))
+        .mapTry(CharSource::read)
+        .getOrElse("");
```
izzyreal, line 16: Maybe deal with exceptions in an exception handler instead of silently swallowing them. You could log them for example.

commons/src/main/java/org/jetbrains/bsp/bazel/commons/BUILD
abrams27: coma is required by formatter

commons/src/main/java/org/jetbrains/bsp/bazel/commons/FileUtils.java
abrams27: I'm working on that

commons/src/main/java/org/jetbrains/bsp/bazel/commons/Constants.java
```diff
@@ -64,5 +64,7 @@
 
   public static final String SCALA_TEST_MAIN_CLASSES_ATTRIBUTE_NAME = "main_class";
 
-  public static final String DEFAULT_PROJECT_VIEW_FILE = "projectview.bazelproject";
+  public static final String PROJECT_VIEW_FILE_PATH = "projectview.bazelproject";
+  public static final String DEFAULT_PROJECT_VIEW_FILE_PATH =
+      ".bazelbsp/default-projectview.bazelproject";
```
izzyreal, line 69: Maybe define `default-projectview.bazelproject` in a central place once and reuse it in `BUILD`, `Install.java` and here.

abrams27: this file in general sucks and entire configuration should be done in a better way. In the following PR project view paths management will be changed anyway, because it will be configurable.

e2e/runTest.sh
```diff
@@ -36,11 +36,16 @@ fi
 bsp_path="$(bazel info bazel-bin)/server/src/main/java/org/jetbrains/bsp/bazel/bsp-install"
 echo "Building done."
 
-echo "Installing BSP..."
+echo "Cleaning project directory..."
 if [ "$#" -eq 2 ]; then
   cd "$TEST_PROJECT_PATH" || exit
 fi
-$bsp_path
+rm -r .bsp/ > /dev/null 2>&1
```
izzyreal, line 43: The `rm` statements still look a little scary, but I don't speak a lot of `sh`. What happens if `[ "$#" -eq 2 ]` is not met? Maybe move the `rm` statements into the `then` scope.

abrams27: by default test is run in the same directory, if statement is only for the case where we want to run test somewhere else

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewSingletonSectionParser.java
```diff
@@ -0,0 +1,31 @@
...
+  public Optional<T> parse(ProjectViewRawSections rawSections) {
+    return rawSections.getLastSectionWithName(sectionName).map(this::parse).flatMap(Try::get);
+  }
+
+  @Override
+  protected Optional<T> parse(String sectionBody) {
+    return Optional.of(sectionBody.strip()).filter(body -> !body.isEmpty()).map(this::instanceOf);
+  }
+
+  protected abstract T instanceOf(String value);
```
izzyreal, line 30: To avoid ambiguity with type-checking (like with Java's `instanceof` operator) I would name this method `instance` or `getInstance`.

Maybe there's a better name for the `value` argument as well, like `sectionBody`.

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewRawSection.java
```diff
@@ -4,16 +4,20 @@
...
-  public ProjectViewRawSection(String sectionHeader, String sectionBody) {
-    this.sectionHeader = sectionHeader;
+  public ProjectViewRawSection(String sectionName, String sectionBody) {
+    this.sectionName = sectionName;
     this.sectionBody = sectionBody;
   }
 
-  public String getSectionHeader() {
-    return sectionHeader;
+  public boolean compareByName(String thatSectionName) {
```
izzyreal, line 15: Maybe unnecessary indirection. Additionally, if it's named `compareByName`, I'd expect the argument to be a `ProjectViewRawSection`.

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewRawSections.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public ProjectViewRawSections(List<ProjectViewRawSection> sections) {
+    this.sections = sections;
+  }
+
+  public Optional<ProjectViewRawSection> getLastSectionWithName(String sectionName) {
+    return getAllWithName(sectionName).stream().reduce((first, second) -> second);
+  }
+
+  public List<ProjectViewRawSection> getAllWithName(String sectionName) {
```
izzyreal, line 21: Maybe return a stream here, since that is what you need in all the non-test consumers of this method anyway.

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewSectionSplitter.java
```diff
@@ -1,91 +1,52 @@
...
 
-  public static List<ProjectViewRawSection> split(String fileContent) {
-    List<ProjectViewSectionHeaderPosition> sectionsHeadersPositions =
-        splitIntoSectionsHeadersPositions(fileContent);
+  private static final String COMMENT_LINE_REGEX = "#(.)*(\\n|\\z)";
+  private static final String COMMENT_LINE_REPLACEMENT = "\n";
 
-    return parseRawSections(fileContent, sectionsHeadersPositions);
-  }
+  public static ProjectViewRawSections split(String fileContent) {
```
izzyreal, line 19:
```suggestion
  public static ProjectViewRawSections getSectionsFromFileContent(String fileContent) {
```

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewSectionSplitter.java
```diff
@@ -1,91 +1,52 @@
...
-    return getStartingAndEndingIndexes(matcher);
+  private static String removeLinesWithComments(String fileContent) {
+    return fileContent.replaceAll(COMMENT_LINE_REGEX, COMMENT_LINE_REPLACEMENT);
   }
 
-  private static List<ProjectViewSectionHeaderPosition> getStartingAndEndingIndexes(
-      Matcher matcher) {
-    ArrayList<ProjectViewSectionHeaderPosition> result = new ArrayList<>();
+  private static List<ProjectViewRawSection> findRawSections(String fileContent) {
+    var sectionHeadersNames = findSectionsHeadersNames(fileContent);
```
izzyreal, line 31:
```suggestion
    String sectionHeadersNames = findSectionsHeadersNames(fileContent);
```

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewSectionSplitter.java
```diff
@@ -1,91 +1,52 @@
...
-  public static List<ProjectViewRawSection> split(String fileContent) {
-    List<ProjectViewSectionHeaderPosition> sectionsHeadersPositions =
-        splitIntoSectionsHeadersPositions(fileContent);
+  private static final String COMMENT_LINE_REGEX = "#(.)*(\\n|\\z)";
+  private static final String COMMENT_LINE_REPLACEMENT = "\n";
 
-    return parseRawSections(fileContent, sectionsHeadersPositions);
-  }
+  public static ProjectViewRawSections split(String fileContent) {
+    var fileContentWithoutComments = removeLinesWithComments(fileContent);
```
izzyreal, line 20: No var. Same for lines 21, 31, 32, 39 and 40.

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewRawSections.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.commons.collections4.CollectionUtils;
+
+public class ProjectViewRawSections {
+
+  private final List<ProjectViewRawSection> sections;
```
izzyreal, line 11:
```suggestion
  private final Stream<ProjectViewRawSection> sections;
```
For extra laziness. Though I'm not sure if streams as class members is a good pattern. But I see the code is using localised streams in quite a few places, so perhaps there are some opportunities to pass these around instead of lists.

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewSingletonSectionParser.java
abrams27: > Maybe there's a better name for the value argument as well, like sectionBody.

well in this scope it is no longer a generic "section body" it is the specific value of a single value section

projectview/src/test/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewSingletonSectionParserTest.java
```diff
@@ -0,0 +1,221 @@
...
+    var expectedSection = sectionConstructor.apply("value with space 123");
+    assertEquals(expectedSection, section.get());
+  }
+
+  // T parse(rawSections)
+
+  @Test
+  public void shouldReturnLastSectionWithoutExplicitDefault() {
+    // given
+    var rawSection1 = new ProjectViewRawSection("anotersection1", "value1");
```
izzyreal, line 128: anoter -> another

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewRawSection.java
abrams27: yeah, it needs to to take `String` so probably a better name would be `.doesHaveName(String)`

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewRawSections.java
abrams27: yea, it think that I should avoid those unnecessary `.collect(Collectors.toList())` cuz I do `.stream()` later anyway

projectview/src/test/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewSingletonSectionParserTest.java
```diff
@@ -0,0 +1,221 @@
...
+    assertFalse(section.isPresent());
+  }
+
+  // T parseOrDefault(rawSections, defaultValue)
+
+  @Test
+  public void shouldReturnLastSection() {
+    // given
+    var rawSection1 = new ProjectViewRawSection("anotersection1", "value1");
+    var rawSection2 = new ProjectViewRawSection(sectionName, "  value2\n");
```
izzyreal, line 168: Looks like the idea is that `sectionName` holds a valid section name, whereas `anotersection*` represent invalid ones. If that's correct, maybe rename these to make that more obvious.

projectview/src/test/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewTargetsSectionParserTest.java
```diff
@@ -0,0 +1,217 @@
...
+    var section = sectionTry.get();
+
+    var expectedSection =
+        new ProjectViewTargetsSection(
+            List.of("//test_included1:test1"),
+            List.of("//test_excluded1:test1", "//test_excluded1:test2"));
+    assertEquals(expectedSection, section);
+  }
+
+  // ProjectViewTargetsSection parse(rawSections)
```
izzyreal, line 120: Maybe a bit more descriptive, so it immediately stands out how these tests differ from the ones that parse a single section.

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/ProjectViewDefaultParserProvider.java
```diff
@@ -1,29 +1,25 @@
...
 
   private static final ProjectViewParser PARSER = new ProjectViewParserImpl();
-  private static final ProjectViewProvider PROJECT_VIEW_PROVIDER = new ProjectViewDefaultProvider();
 
   private final Path projectViewFile;
+  private final Path defaultProjectViewFile;
 
   public ProjectViewDefaultParserProvider(Path bspProjectRoot) {
-    this.projectViewFile = bspProjectRoot.resolve(Constants.DEFAULT_PROJECT_VIEW_FILE);
+    this.projectViewFile = bspProjectRoot.resolve(Constants.PROJECT_VIEW_FILE_PATH);
```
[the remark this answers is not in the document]
abrams27: tbh I love mapping constructor body -> always `this.<...>` it is annoying to see some `this` in some constructors and not in some

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewSectionSplitter.java
abrams27: why?

projectview/src/test/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewSingletonSectionParserTest.java
abrams27: `anotersection*` represents another section, it can be valid (it will be parsed by some parser) but it is not the one we are looking for

projectview/src/test/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewTargetsSectionParserTest.java
abrams27: I want to bump JUnit to version 5, then I will split it into nested tests, with better description

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewSingletonSectionParser.java
izzyreal: wrt "value", the only code I could find that consumes `instanceOf` is:
```
  protected Optional<T> parse(String sectionBody) {
    return Optional.of(sectionBody.strip()).filter(body -> !body.isEmpty()).map(this::instanceOf);
  }
```
I suppose there's a fine line between "section" and "value"?

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewSectionSplitter.java
izzyreal: So it's clear what the types are. Probably not an issue for lines 21 and 39 though, since their names give away (something clear) about the types.

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewRawSections.java
izzyreal: Ah of course...

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/splitter/ProjectViewSectionSplitter.java
abrams27: well, IntelliJ says that in a very pretty way

server/src/main/java/org/jetbrains/bsp/bazel/server/ServerInitializer.java
```diff
@@ -68,20 +68,23 @@ public static void main(String[] args) {
   }
 
   private static BazelBspServerConfig getBazelBspServerConfig(Path bspProjectRoot, String[] args) {
-    var pathToBazel = args[0];
     var projectView = getProjectView(bspProjectRoot, args);
+    var pathToBazel = projectView.getBazelPath().get().getValue();
```
izzyreal, line 72: `.get()` is ok here? Also line 79.

abrams27: yea, if it is empty, test will fail anyway (I remember that there was some implicit for it in scala)

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/config/ServerArgsProjectViewProvider.java
```diff
@@ -1,28 +1,47 @@
...
+import java.util.Optional;
 import org.jetbrains.bsp.bazel.projectview.model.ProjectView;
 import org.jetbrains.bsp.bazel.projectview.model.ProjectViewProvider;
-import org.jetbrains.bsp.bazel.projectview.model.sections.specific.DirectoriesSection;
-import org.jetbrains.bsp.bazel.projectview.model.sections.specific.TargetsSection;
+import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewBazelPathSection;
+import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewTargetsSection;
+import org.jetbrains.bsp.bazel.projectview.parser.ProjectViewDefaultParserProvider;
 
+@Deprecated
```
izzyreal, line 14: Is there an alternative? Currently all implementations of `ProjectViewProvider` are deprecated.

projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/parser/sections/ProjectViewSingletonSectionParser.java
abrams27: section contains value (or list of values) and a name

PR #151, opened 2022-02-16
Store BazelProcess output inside BazelProcessResult

### LINKED ISSUE #150: BazelProcess.waitAndGetResult() always returns empty stdout and stderr
This happens because the contents of these streams are [drained](https://github.com/JetBrains/bazel-bsp/blob/8773ecd475a658c20d609f6efc08619cd73d8173/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.java#L28) by the logging method.

odisseus:
Fixes #150.

### DIFF
--- a/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/data/BazelProcessResult.java
+++ b/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/data/BazelProcessResult.java
@@ -3,20 +3,23 @@ package org.jetbrains.bsp.bazel.bazelrunner.data;
 import ch.epfl.scala.bsp4j.StatusCode;
 import java.io.InputStream;
 import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
 import org.jetbrains.bsp.bazel.bazelrunner.utils.BazelStreamReader;
 import org.jetbrains.bsp.bazel.commons.ExitCodeMapper;
+import org.jetbrains.bsp.bazel.commons.Lazy;
 
 public class BazelProcessResult {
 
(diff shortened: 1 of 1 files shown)


abrams27 (commented):
Laziness is cool, but I feel that here we can just drain the Stream in the constructor and just store `List<String>` in the class

abrams27:
also please change `dev` branch to `master` - we are getting rid of a `dev` branch

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/data/BazelProcessResult.java
```diff
@@ -25,11 +28,35 @@ public StatusCode getStatusCode() {
...
+    return stdout.getValue().get();
   }
 
   public String getJoinedStderr() {
-    List<String> lines = BazelStreamReader.drainStream(stderr);
+    List<String> lines = stderr.getValue().get();
     return String.join(LINES_DELIMITER, lines);
   }
+
+  private static class StreamStorage extends Lazy<List<String>> {
```
lukaszwawrzyk, line 39: for similar use case I used Suppliers.memoize.

```
private Supplier<String> memoizedOutput(InputStream input) {
  return Suppliers.memoize(() -> BazelStreamReader.drainStream(input));
}

```

lukaszwawrzyk:
> Laziness is cool, but I feel that here we can just drain the Stream in the constructor and just store `List<String>` in the class

Laziness might be useful, if we don't care about the output (and sometimes we don't), there is no point to do the IO to read it.

odisseus:
See #154.

PR #154, opened 2022-02-16
Memoize BazelProcess output so that it doesn't get lost

### LINKED ISSUE #150: BazelProcess.waitAndGetResult() always returns empty stdout and stderr

odisseus:
Fixes #150.


abrams27 (approved):
looks great, thanks! (please don't merge yet @lukaszwawrzyk - I'm doing stuff around master branch)

PR #159, opened 2022-02-17
[change] e2e tests are now testing all major versions of bazel

abrams27:
in the near future we will be moving from GH actions to Team City, and then I will make CI configuration better, but for now it should be sufficient

**NOTE** probably CI configuration will be updated after bazel configuration changes

---

fixes https://github.com/JetBrains/bazel-bsp/issues/153

PR #160, opened 2022-02-17
[change] `5.0.0` bazel version

### LINKED ISSUE #140: latest version of bazel breaks the build
latest versions of Bazel (> `3.7.2`) break the build

abrams27:
the latest bazel version!

I decided that it would be nice to use `5.0.0` bazel version in order to try out the latest possible version of bazel, to notice any needed changes to our server (dogfooding for the win)

I will deal with e2e tests (bazel versions) in https://github.com/JetBrains/bazel-bsp/pull/159

---
fixes #140 

- [x] blocked by https://github.com/JetBrains/bazel-sonatype/pull/1 - I would like to use it with proper version tag

---
testing
```
bazel build //...
bazel test //...
```


lukaszwawrzyk:
I learned couple of times that it is better to have pinned versions of everything to have some determinism and reproducibility.
Imagine that you want to push some PR (possibly quickly) and you push a cosmetic change and it starts to fail and  you are debugging it for ages, it may even behave different on CI and locally. This can easily be caused by a new bazel release. In fact we already had it once when the version not set at all and depended on what is installed on given machine.
A better approach is to have pinned version that is updated automatically by a bot that opens a PR with a bump. If CI starts failing on such PR it means that the problem is connected to bazel version change. At the same time all other PRs can proceed and someone can work specifically on that version bump. 
I am not sure how much work it is to set up automatic bumps, if much, I would stick to pinned version updated manually.

abrams27:
> I learned couple of times that it is better to have pinned versions of everything to have some determinism and reproducibility.
> Imagine that you want to push some PR (possibly quickly) and you push a cosmetic change and it starts to fail and you are debugging it for ages, it may even behave different on CI and locally. This can easily be caused by a new bazel release. In fact we already had it once when the version not set at all and depended on what is installed on given machine.
> A better approach is to have pinned version that is updated automatically by a bot that opens a PR with a bump. If CI starts failing on such PR it means that the problem is connected to bazel version change. At the same time all other PRs can proceed and someone can work specifically on that version bump.
> I am not sure how much work it is to set up automatic bumps, if much, I would stick to pinned version updated manually.

okey, so `5.0.0`?

lukaszwawrzyk:
5.0.0 is ok, I am also fine with any newer version, as long as we have tests for 5.0.0.

PR #165, opened 2022-02-23
fix

abrams27:


lukaszwawrzyk (approved):
Fine as a hotfix, as it is going to be replaced anyway.

Issue #144, opened 2022-02-11
project view parser needs logging


abrams27:
we need more logs from project view parser

Issue #158, opened 2022-02-16
workspace context


abrams27:
I think that it would be nice to have a `workspace context` (temporary name) to keep all information about the workspace - targets, path to root, bazelrunner all information needed during server lifetime

also, this would be very useful in terms project view usage, I think that project view should be mapped into workspece context, since project view might contain redundant information - for example we don't need java path used during initialisation or debugger address. Also, it might be useful during `derive_targets_from_directories` implementation - targets inside workspace context would be deducted on provided targets / directories + `derive_targets_from_directories` flag

lukaszwawrzyk:
It sounds useful. I plan to have something similar, maybe a bit smaller in scope though. I have a very early version of such thing. It is Project together with ProjectStorage. I want to be able to ask ProjectStorage for Project instance at any point. The Project would essentially contain all information from sync (output from aspect(s?)). Currently it contains the resolved targets that should be imported as well as TargetInfo for each target.
I would like it to contain proper project structure like modules with all relevant attributes like sources, dependencies, SDKs.
The idea is that workspaceBuildTargets triggers a sync, which calls bazel and builds the project from the beginning, and then stores it in the ProjectStorage. Then, any other endpoint (including workspaceBuildTargets) can ask ProjectStorage for Project, map its contents to BSP protocol classes and send the response. 
ProjectStorage should also save the project to disk in some format and be able to read it later, so that when server is restarted it doesn't need to sync again to work. We may also consider storing some metadata to support incremental sync.

Issue #120, opened 2021-12-15
Project views directory inference doesn't seem to work


lukaszwawrzyk:
It is not documented enough. I didn't know where to put the `projectview.bazelproject`, I hoped it would be .bsp dir, but it seems to be at repository root. It is hard to figure out from code. Also I found some undocumented way of passing targets through command line (?). Not sure how to use it. 

I tried with following content:

```yaml
directories:
  server/src/main/java/org/jetbrains/bsp/bazel/server/bep

derive_targets_from_directories: true

targets:

additional_languages:
  kotlin
  python
  scala
```

Nothing was imported. I had to add targets as well:

```yaml
directories:
  server/src/main/java/org/jetbrains/bsp/bazel/server/bep

derive_targets_from_directories: true

targets:
  //server/src/main/java/org/jetbrains/bsp/bazel/server/bep/...:all

additional_languages:
  kotlin
  python
  scala
```

abrams27:
> It is not documented enough. 

i'll create a documentation

> I didn't know where to put the projectview.bazelproject, I hoped it would be .bsp dir, but it seems to be at repository root. It is hard to figure out from code. Also I found some undocumented way of passing targets through command line (?). Not sure how to use it.

i want to change it - projectview will be loaded also during a `workspace/reload` call, so it would be nice to configure which projectview file should be loaded, but I'm not sure how to do that yet


> I tried with following content:
> 
> ```yaml
> directories:
>   server/src/main/java/org/jetbrains/bsp/bazel/server/bep
> 
> derive_targets_from_directories: true
> 
> targets:
> 
> additional_languages:
>   kotlin
>   python
>   scala
> ```
> 
> Nothing was imported. I had to add targets as well:
> 
> ```yaml
> directories:
>   server/src/main/java/org/jetbrains/bsp/bazel/server/bep
> 
> derive_targets_from_directories: true
> 
> targets:
>   //server/src/main/java/org/jetbrains/bsp/bazel/server/bep/...:all
> 
> additional_languages:
>   kotlin
>   python
>   scala
> ```

`derive_targets_from_directories: true` is not supported yet (again, lack of documentation, sorryy)

lukaszwawrzyk:
> i want to change it - projectview will be loaded also during a `workspace/reload` call, so it would be nice to configure which projectview file should be loaded, but I'm not sure how to do that yet

[...]

PR #166, opened 2022-02-24
[change] project view parser has more logging, targets section is optional and sections are using specific types instead of raw strings

abrams27:
introducing:
* more logs (closes https://github.com/JetBrains/bazel-bsp/issues/144)
* `targets` section is now optional - preparation for https://github.com/JetBrains/bazel-bsp/issues/158 and https://github.com/JetBrains/bazel-bsp/issues/120)
* sections are using specific types instead of raw strings


projectview/src/main/java/org/jetbrains/bsp/bazel/projectview/model/ProjectView.java
```diff
@@ -161,18 +162,34 @@ private ProjectViewTargetsSection combineTargetsSection(
...
+        List<V> combineListValuesWithImported(
+            List<ProjectView> importedProjectViews,
+            T section,
+            Function<ProjectView, T> sectionGetter,
+            Function<S, List<V>> valuesGetter) {
+      var sectionValues = section.map(valuesGetter).orElse(List.of());
+
       return importedProjectViews.stream()
           .map(sectionGetter)
+          .filter(Optional::isPresent)
```
lukaszwawrzyk, line 179: is `flatMap(Optional::stream)` on this java version?

lukaszwawrzyk (approved):

PR #182, opened 2022-03-07
Improve build client logging

### LINKED ISSUE #168: If `bazel info exec_root` fails contents of stderr should be printed

agluszak:
Log error if bazel data cannot be resolved.
Make build client logging more explicit.
Fixes #168

### DIFF
--- a/bazelrunner/BUILD
+++ b/bazelrunner/BUILD
@@ -9,6 +9,14 @@ java_library(
 )
 
 java_library(
+    name = "utils",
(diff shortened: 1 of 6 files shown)


agluszak:
> I feel that it will conflict with #147, so maybe we can postpone it?

sure

abrams27 (changes requested):
please rebase, maybe the change is no longer needed

abrams27:
is it still needed? @agluszak

PR #188, opened 2022-03-14
[change] improve source dependencies performance

lukaszwawrzyk:
Traversing dependency graph the way it was implemented (very naively) is painfully slow, which can be noticed on bigger projects. It would be nice to have a graph built once and just take all direct deps from it based on condition to not duplicate them.
As a fairly reasonable workaround, I just take transitive source deps directly from bazel where they are already computed anyway. BSP client will deduplicate and group it anyway.


abrams27 (approved):
the formatting fails + please update changelog (with sync mechanism PR)

PR #198, opened 2022-03-18
[feature] output forwarding from bazel processes

lukaszwawrzyk:
Now, each bazel process run with `BazelRunner` (when `waitAndGetResult()` is called) will send log message to bsp-client:
Running: [command] and forward stdout and stderr. The output is also collected to string and lines list for later use.
PR also includes some refactoring to vavr, refactoring of bazel flags etc. 

I also replaced includedTargets and excludedTargets pair with single class TargetSpecs. In the future I imagine it as the main way to specify what should be built/run. It can also contain directories to derive targets from. Bazel runner could resolve what to run and run it.


abrams27 (commented):
looks good, but the test fails - I couldn't spot the reason

abrams27:
ohh and I thing that it would be nice to color (or make it bold if it is possible) `invoking: ...` line as well

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.java
```diff
@@ -1,61 +1,34 @@
...
     this.process = process;
-    this.buildClientLogger = buildClientLogger;
+    this.logger = logger;
   }
 
   public BazelProcessResult waitAndGetResult() {
     try {
+      var stdoutCollector = new OutputCollector();
+      var stderrCollector = new OutputCollector();
+      AsyncOutputProcessor.run(process.getInputStream(), stdoutCollector, logger::logMessage);
```
abrams27, line 22: we can add logging to file here as well

lukaszwawrzyk:
> looks good, but the test fails - I couldn't spot the reason

I think I have a race condition. Working on a fix.

lukaszwawrzyk:
> ohh and I thing that it would be nice to color (or make it bold if it is possible) `invoking: ...` line as well

The problem is that the sync console displays color but the build console doesn't (and you see some unprintable characters). So I just carefully enabled colors only on calls when it is useful and displays in sync console. If I colored the Invoking it would not always display correctly :/

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.java
abrams27: yep

abrams27 (approved):
please update changelog + logging to file

CHANGELOG.md
```diff
@@ -20,6 +21,9 @@
 
 ### Changes 🔄
 
+- Improved bazel runner. The BSP Client now receives invoked bazel command,
```
abrams27, line 24: I think that we can consider it as a feature

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/utils/Stopwatch.java
```diff
@@ -0,0 +1,23 @@
+package org.jetbrains.bsp.bazel.bazelrunner.utils;
+
+import java.time.Duration;
+
+public class Stopwatch {
```
abrams27, line 5: u can use guava stop watch as well if u want to: https://guava.dev/releases/24.0-jre/api/docs/com/google/common/base/Stopwatch.html

Issue #205, opened 2022-03-29
Server state is not maintained between bazel-bsp runs


lukaszwawrzyk:
Server state is not maintained between bazel-bsp runs. As a consequence full sync is needed to run tests after IDE restart.
My idea is to simply store `Project` as json and read it on startup if exists to recover in memory state.

PR #207, opened 2022-03-31
[feature] project storage mechanism

lukaszwawrzyk:
I implemented ProjectStorage to store `Project` between runs.

I introduced `bspClientLogger.timed("Operation", () -> {})` to time and report some long operations.

I refactored BazelData to be BazelInfo (as the output of bazel info) and created BspInfo as it is messy to have everything in one object and introduces unnecessary dependencies.

I renamed BuildClientLogger -> BspClientLogger. I think it will be more clear now that it logs to the bsp client. 

There is one more change here, could be a separate PR but I was lazy (sorry).
In case of aborted event from BEP which happens when some BUILD files are incorrect for example, I send a warning to bsp client (also client sees build output that explains the error precisely). With the exception approach, client sees a lot of ugly (new line between each actual line) stacktraces (the stacktrace doesn't help with anything, it is just form bep). Not sure why this exception doesn't interrupt the import, but it shouldn't anyway. We want to import a project at least partially even if it doesn't build.

---

fixes https://github.com/JetBrains/bazel-bsp/issues/205


server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaModule.java
```diff
@@ -107,4 +110,20 @@ public int hashCode() {
         sourcesClasspath,
         ideClasspath);
   }
+
+  @Override
+  public String toString() {
```
abrams27, line 115: this this could be generated using intellij, right?

abrams27 (approved):
wow, I'm loving it!

abrams27:
there is only one issue - if user changes the project view and the cache is still there, the change from project view won't be applied, am I right? also I couldn't see the message when I reloaded the project / imported again "Loading project from local cache ...." - when should I see it?

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaModule.java
lukaszwawrzyk: Default intellij generated to string is not readable enough if you want to actually understand something. This approach is farily custom, I specify how to render some sequences (long or short). Maybe it could be automated more.

lukaszwawrzyk:
> there is only one issue - if user changes the project view and the cache is still there, the change from project view won't be applied, am I right?

I was thinking about this and IMO this is a separate mechanism. Generally sync should be triggered manually (optionally user may want to enable auto sync if project view changes, etc.). You may imagine that we should detect on startup if git commit or unstaged changes changed as well as project view and sync project instead of loading it from cache. But if so, we should be also doing this if intellij is running, and we are not. We don't want to trigger possibly expansive sync when someone changed branch for a minute and will go back, or if someone is happy with current (possibly incorrect, but possibly still correct enough) state of the project.

> also I couldn't see the message when I reloaded the project / imported again "Loading project from local cache ...." - when should I see it?

You will see it only when you sync, kill the server/restart IDE and run tests/app. It will load project from cache and run bazel info (i am considering to similarly cache bazel info in the future). You will not see it when you just trigger another sync with this file existing. If you click sync, it means you don't want to use cache, you want to reload everything from bazel.

PR #223, opened 2022-04-20
[change] logging level has been set to info and doesn't include events

abrams27:
- logging level = info
- there is a single logger configuration, not multiple for each module (e2e still needs different logging)
- bep events are now `debug` not `info`

---

https://youtrack.jetbrains.com/issue/BAZEL-15

---

testing:

manual testing


server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepServer.java
```diff
@@ -177,7 +177,7 @@ private void consumeActionEvent(BuildEventStreamProtos.ActionExecuted action) {
     Map<Uri, List<PublishDiagnosticsParams>> filesToDiagnostics =
         action.getActionMetadataLogsList().stream()
             .filter(log -> log.getName().equals(Constants.DIAGNOSTICS))
-            .peek(log -> LOGGER.info("Found diagnostics file in {}", log.getUri()))
+            .peek(log -> LOGGER.debug("Found diagnostics file in {}", log.getUri()))
```
lukaszwawrzyk, line 180: will conflict with diagnostics

lukaszwawrzyk (approved):

server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepServer.java
abrams27: i'll merge it after diagnostics

PR #225, opened 2022-04-20
[feature] new diagnostics

lukaszwawrzyk:
New implementation of diagnostics extraction. It collects both warnings and errors, doesn't show irrelevant errors in BUILD files, shows line and character position and does not run bazel query before each build anymore


server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/BazelRootMessageParser.kt
```diff
@@ -0,0 +1,66 @@
+package org.jetbrains.bsp.bazel.server.diagnostics
+
+object BazelRootMessageParser : Parser {
+  private const val TargetLabel = """(//[\w/.-]+:[\w/.-]+)"""
```
abrams27, line 4: it wont match targets like `//:target`

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/BazelRootMessageParser.kt
```diff
@@ -0,0 +1,66 @@
...
+      .*              # part of actual error message
+      $               # end of line
+      """.toRegex(RegexOption.COMMENTS)
+
+  private fun findErrorInBUILD(output: Output): List<Diagnostic>? {
+    return output.tryTake(ErrorInBUILD)
+        ?.let { match ->
+          val targetLabel = match.groupValues[4].ifEmpty { null }
+          val compilerDiagnostics = collectCompilerDiagnostics(output, targetLabel)
+          if (compilerDiagnostics.isEmpty()) {
```
abrams27, line 31: could be replaced with `ifEmpty`

abrams27 (commented):
wow!

abrams27 (approved):
lets gooo

PR #233, opened 2022-04-22
[change] do not serialize label->module map

lukaszwawrzyk:
Automagic serialization was doubling size of the file by serializing field that is derived from the actual state in the constructor.


abrams27 (approved):
changelog plix

PR #238, opened 2022-05-05
[change] diagnostic message doesn't contain path to file anymore & column number is now inferred from the message | #BAZEL-51 fixed

abrams27:
* diagnostic message doesnt contain path to file - intellij handles the hyperlinks
* for most of cases column number is inferred from number of spaces before `^` sign 
(see tests)

---

https://youtrack.jetbrains.com/issue/BAZEL-51

---
```
bazel test //server/src/test/java/org/jetbrains/bsp/bazel/server/diagnostics:DiagnosticsServiceTest
```

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/CompilerDiagnosticParser.kt
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/CompilerDiagnosticParser.kt
@@ -23,11 +23,16 @@ object CompilerDiagnosticParser : Parser {
           ?.let { match ->
             val path = match.groupValues[1]
             val line = match.groupValues[2].toInt()
-            val column = match.groupValues[3].toIntOrNull() ?: 1
+            val column = match.groupValues[3].toIntOrNull() ?: tryFindColumnNumber(output) ?: 1
             val level = if (match.groupValues[4] == "warning") Level.Warning else Level.Error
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/CompilerDiagnosticParser.kt
```diff
@@ -23,12 +23,17 @@ object CompilerDiagnosticParser : Parser {
...
-            val column = match.groupValues[3].toIntOrNull() ?: 1
+            val column = match.groupValues[3].toIntOrNull() ?: tryFindColumnNumber(output) ?: 1
             val level = if (match.groupValues[4] == "warning") Level.Warning else Level.Error
-            val message = constructMessage(match.value, output)
+            val message = constructMessage(match.groupValues[5], output)
             Diagnostic(Position(line, column), message, level, path, targetLabel)
           }
 
+  private fun tryFindColumnNumber(output: Output): Int? {
+    val line = output.peek(limit = 20).find { IssuePositionMarker.matches(it) }
```
lukaszwawrzyk, line 33: Maybe it would be cleaner to take `message` and look for `^` there? In theory you can peek something else here and you are skipping the logic of constructMessage which can change for example.

lukaszwawrzyk (approved):

PR #244, opened 2022-05-10
[change] various refactoring and performance improvements

steveniemitz:
This is an initial refactoring to set up for the next review to add bloop support.  There's a few refactorings in here to pull out some things into base classes (that'll be used elsewhere) and also a couple bug fixes/performance enhancements.

Notably:
- We use a repository override to map the aspect into the project we're inspecting.  This has the benefit that we no longer need the aspect to be inside the project.
- We use the Action completed event to get the (failed) output for targets.  This removes the need to parse most of the output and makes it a little more robust
- Anywhere that used `vavr.List` has been changed to `vavr.Seq`.  The underlying implementation is now generally an Array rather than a linked list.  This substantially reduced allocations and improved performance.


install/src/main/resources/aspects.bzl
```diff
@@ -307,6 +307,10 @@ def extract_java_runtime(target, ctx, dep_targets):
 
     if java_common.JavaRuntimeInfo in target:
         runtime = target[java_common.JavaRuntimeInfo]
+    else:
```
lukaszwawrzyk, line 310: this part is handled on my current PR

server/src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -39,23 +42,63 @@ public class BazelBspServer {
...
+
+  public static ServerContainer makeProjectProvider(
+      BspInfo bspInfo,
+      WorkspaceContextProvider workspaceContextProvider,
+      Option<Path> workspaceRoot,
+      Option<ProjectStorage> projectStorage) {
+    var bspClientLogger = new BspClientLogger();
+    var bazelInfoStorage = new BazelInfoStorage(bspInfo);
+    var bazelDataResolver =
+        workspaceRoot
```
lukaszwawrzyk, line 86: Not sure what happens here. The idea of bazel runner constructors was that `inCwd` just runs bazel to resolve bazel info (and it is assumed that it is run somewhere inside bazel workspace), so it is used at most once, to construct proper bazel runner with workspace path that was extracted from bazel. IF now we have a better way to find workspace we can drop this behavior.

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspAspectsManager.java
```diff
@@ -19,15 +21,17 @@ public BazelBspAspectsManager(
...
             targetSpecs,
-            List.of(
+            Array.of(
+                BazelFlag.repositoryOverride(
+                    Constants.ASPECT_REPOSITORY, aspectsResolver.getBazelBspRoot()),
                 BazelFlag.aspect(aspectsResolver.resolveLabel(aspect)),
                 BazelFlag.outputGroups(outputGroups.toJavaList()),
                 BazelFlag.keepGoing(),
-                BazelFlag.color(true)));
+                BazelFlag.color(false)));
```
lukaszwawrzyk, line 34: :/

install/src/main/resources/aspects.bzl
steveniemitz: :+1: I can revert this.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ModuleMerger.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.jetbrains.bsp.bazel.server.sync.languages.LanguageData;
+import org.jetbrains.bsp.bazel.server.sync.languages.java.JavaModule;
+import org.jetbrains.bsp.bazel.server.sync.languages.scala.ScalaModule;
+import org.jetbrains.bsp.bazel.server.sync.model.Label;
+import org.jetbrains.bsp.bazel.server.sync.model.Module;
+import org.jetbrains.bsp.bazel.server.sync.model.SourceSet;
+
+public class ModuleMerger {
```
lukaszwawrzyk, line 15: Why do we need this? IntelliJ gets incompatible modules from bsp and merges them on its own.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/TargetKindResolver.java
```diff
@@ -7,22 +7,28 @@
...
     var tag =
         ruleSuffixToTargetType.entrySet().stream()
             .filter(entry -> targetInfo.getKind().endsWith("_" + entry.getKey()))
             .map(Map.Entry::getValue)
             .findFirst()
-            .orElse(Tag.NO_IDE);
+            .orElse(NO_IDE);
     if (targetInfo.getTagsList().contains("no-ide")) {
-      return HashSet.of(tag, Tag.NO_IDE);
+      return NO_IDE;
```
lukaszwawrzyk, line 30: Is it safe to drop tag? AFAIR IntelliJ is still processing such modules somehow, just not creating them in the end.

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticsService.kt
```diff
@@ -1,16 +0,0 @@
-package org.jetbrains.bsp.bazel.server.diagnostics
-
-import ch.epfl.scala.bsp4j.PublishDiagnosticsParams
-import org.jetbrains.bsp.bazel.bazelrunner.BazelInfo
-
-class DiagnosticsService(bazelInfo: BazelInfo) {
```
lukaszwawrzyk, line 6: I would keep this class. We have ProjectSyncService and ExecutionService that handle bsp structures (inputs bsp and outputs bsp). They are responsible for mapping inputs, calling some internal independent api and map outputs. I wanted to keep this convention with diagnostics service. It is a small class, but just keeps things clean and consistent.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaLanguagePlugin.java
```diff
@@ -100,14 +108,14 @@ protected void applyModuleData(JavaModule javaModule, BuildTarget buildTarget) {
 
   public JvmBuildTarget toJvmBuildTarget(JavaModule javaModule) {
     var jdk = javaModule.jdk();
-    var javaHome = jdk.javaHome().map(URI::toString).getOrNull();
+    var javaHome = jdk.javaHome().map(Path::toString).getOrNull();
```
abrams27, line 111: BSP expects uri not path

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaLanguagePlugin.java
```diff
@@ -100,14 +108,14 @@ protected void applyModuleData(JavaModule javaModule, BuildTarget buildTarget) {
...
     return new JvmBuildTarget(javaHome, jdk.javaVersion());
   }
 
   public JvmEnvironmentItem toJvmEnvironmentItem(Module module, JavaModule javaModule) {
     return new JvmEnvironmentItem(
         toBspId(module),
-        javaModule.runtimeClasspath().map(URI::toString).asJava(),
+        javaModule.runtimeClasspath().map(Path::toString).asJava(),
```
abrams27, line 118: BSP expects uri not path

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaLanguagePlugin.java
```diff
@@ -117,7 +125,7 @@ public JavacOptionsItem toJavacOptionsItem(Module module, JavaModule javaModule)
     return new JavacOptionsItem(
         toBspId(module),
         javaModule.javacOpts().toJavaList(),
-        javaModule.ideClasspath().map(URI::toString).toJavaList(),
+        javaModule.ideClasspath().map(Path::toString).toJavaList(),
```
abrams27, line 128: BSP expects uri not path - import with intellij fails

server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepServer.java
```diff
@@ -127,6 +127,36 @@ private void processCompletedEvent(BuildEventStreamProtos.BuildEvent event) {
     }
   }
 
+  private void processActionCompletedEvent(BuildEventStreamProtos.BuildEvent event) {
+    if (event.getId().hasActionCompleted()) {
+      consumeActionCompletedEvent(event);
+    }
+  }
+
+  private void consumeActionCompletedEvent(BuildEventStreamProtos.BuildEvent event) {
```
abrams27, line 136: i believe that this method could be split (if - else section is kinda messy)

abrams27 (changes requested):
thanks for the PR!

bsp expects uri almost everywhere, and u missed `.toURI` in few places (eg. classpath, javaHome) , so the import fails - u can try it on your own using `install.sh` script

also, we were thinking about changing the way how `argv` argument in the configuration file looks - we want to use coursier there (https://youtrack.jetbrains.com/issue/BAZEL-63) - would it be a problem for the bloop case?

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ModuleMerger.java
abrams27: + in the future the client will handle such case even better

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/TargetKindResolver.java
abrams27: it doesnt since we are filtering it out in `BazelProjectMapper#createModules`

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ModuleMerger.java
steveniemitz: I've had a few people say this now (that IJ handles merging modules), in practice I've rarely seen it actually work correctly.  Can you point me to where it does this?

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticsService.kt
steveniemitz: yeah that's cool, I can add this back.

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticBspMapper.kt
```diff
@@ -12,24 +12,22 @@ import ch.epfl.scala.bsp4j.Range as BspRange
...
-      fileDiagnostics: List<Diagnostic>
-  ): PublishDiagnosticsParams {
-    val fileLocation = TextDocumentIdentifier(toAbsoluteUri(rawFileLocation))
-    val targetId = BuildTargetIdentifier(fileDiagnostics[0].targetLabel ?: "")
-    val diagnostics = fileDiagnostics.map { createDiagnostic(it) }
-    return PublishDiagnosticsParams(fileLocation, targetId, diagnostics, false)
+  fun createDiagnostics(diagnostics: List<Diagnostic>, target: String): List<PublishDiagnosticsParams> {
+    val ident = BuildTargetIdentifier(target)
+    if (diagnostics.isEmpty()) {
+      return listOf(PublishDiagnosticsParams(TextDocumentIdentifier("unknown"), ident, listOf(), false))
```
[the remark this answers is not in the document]
steveniemitz: hm?

steveniemitz:
> bsp expects uri almost everywhere, and u missed `.toURI` in few places (eg. classpath, javaHome) , so the import fails - u can try it on your own using `install.sh` script

Yeah, actually after more thought I'm going to move everything back to using URI.  Keeping it as a path optimizes for my specific case that I benchmarked (refreshing the model and exporting it to bloop) but will push the cost down to every BSP operation instead, which imo isn't worth it.  We can probably optimize Path -> URI instead.

> also, we were thinking about changing the way how `argv` argument in the configuration file looks - we want to use coursier there (https://youtrack.jetbrains.com/issue/BAZEL-63) - would it be a problem for the bloop case?

This should be fine.  We're using coursier to launch bloop too actually.

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticBspMapper.kt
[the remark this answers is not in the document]
abrams27: i mean why we would like to send such diagnostics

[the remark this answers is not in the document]
steveniemitz: oh, do you mean we should just return an empty list here instead?  I figured this way we'd at least know a target failed even if we can't figure out why.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/TargetKindResolver.java
lukaszwawrzyk: True that, but I am not sure if this is valid in terms of protocol. I just did it 'temporarly' having in mind that we might revert it if it causes any problems and solve it properly. Anyway if it happens, we can adjust this place as well.

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticBspMapper.kt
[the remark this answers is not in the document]
lukaszwawrzyk: I would keep the change, we will observe it in practice, if it even displays and if is useful and then decide.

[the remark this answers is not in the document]
abrams27: okey, then imo it should be inside `AllCatchParser`, not in the mapper

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ModuleMerger.java
abrams27: it's inside current bsp client (https://github.com/JetBrains/intellij-scala/tree/idea221.x/bsp) - it creates a single IJ module using multiple bsp modules - it just merges all the sources / opts / names and creates single IJ module. We are slowly working on a new client, with better support for that

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticBspMapper.kt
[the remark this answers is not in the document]
steveniemitz: I split the difference here and added it back to the DiagnosticsService, it's a little tricky to plumb it into the AllCatchParser since it'd need the target name there.

server/src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -41,23 +44,63 @@ public class BazelBspServer {
...
+    this.bspServerApi =
+        new BspServerApi(
+            serverLifetime,
+            bspRequestsRunner,
+            projectSyncService,
+            executeService,
+            cppBuildServerService);
+  }
+
+  public static ServerContainer makeProjectProvider(
```
abrams27, line 80: it can be private

abrams27, line 80: it can be private

server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepServer.java
```diff
@@ -127,6 +126,39 @@ private void processCompletedEvent(BuildEventStreamProtos.BuildEvent event) {
...
+    var label = event.getId().getActionCompleted().getLabel();
+    var actionEvent = event.getAction();
+    if (!actionEvent.getSuccess()) {
+      consumeUnsuccessfulActionCompletedEvent(actionEvent, label);
+    }
+  }
+
+  private void consumeUnsuccessfulActionCompletedEvent(
+      BuildEventStreamProtos.ActionExecuted actionEvent, String label) {
+    String stdErrText = "";
```
abrams27, line 145: i meant that instead of changing stdErrText and using it in last two lines, u can extract two last lines to a method and just call it in those ifs

abrams27, line 145: i meant that instead of changing stdErrText and using it in last two lines, u can extract two last lines to a method and just call it in those ifs

server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepServer.java
```diff
@@ -127,6 +126,39 @@ private void processCompletedEvent(BuildEventStreamProtos.BuildEvent event) {
...
+    if (actionEvent.getStderr().getFileCase() == BuildEventStreamProtos.File.FileCase.URI) {
+      var path = Paths.get(URI.create(actionEvent.getStderr().getUri()));
+      try {
+        stdErrText = new String(Files.readAllBytes(path), StandardCharsets.UTF_8);
```
abrams27, line 149: could be replaced with `Files.readString()`

abrams27, line 149: could be replaced with `Files.readString()`

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ModuleMerger.java
abrams27: bump i believe that we dont want to do it here - it's kinda agains protocol, it might have strange conflicts with the clients mechanism + new IJ BSP client solves it

abrams27 (changes requested):

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticBspMapper.kt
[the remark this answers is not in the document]
abrams27: >I split the difference here and added it back to the DiagnosticsService, it's a little tricky to plumb it into the AllCatchParser since it'd need the target name there.

we could pass target name to parsers, then we would be able to set target name for java compiler

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ModuleMerger.java
lukaszwawrzyk: We should make projectviews more flexible to put there and read config easily. Perhaps add a generic feature flags (so that we can do for example `bazel-bsp -t //my/target/... -f merge-modules=true -f add-synthetic-module-roots=true`. Such feature would be opt in. If it really makes bazel-bsp work better in intellij right now, I would take it. There is similar situation with my synthetic modules. They should be behind a flag. What should be the default is up for discussion.

server/src/main/java/org/jetbrains/bsp/bazel/server/BazelBspServer.java
```diff
@@ -109,12 +149,36 @@ private void integrateBsp(BspIntegrationData bspIntegrationData) {
...
+    var bepServer = new BepServer(client, new DiagnosticsService(bazelInfo));
     compilationManager.setBepServer(bepServer);
     bspIntegrationData.setServer(ServerBuilder.forPort(0).addService(bepServer).build());
   }
 
   public void setBesBackendPort(int port) {
     bazelRunner.setBesBackendPort(port);
   }
+
+  public static class ServerContainer {
```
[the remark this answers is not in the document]
steveniemitz: this'll be used later in the bloop exporter, so it needs to be public

server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepServer.java
steveniemitz: I ran into issues using methods that were introduced in java 9+, are we targeting only 9+ or maintaining compatibility with 8?

steveniemitz: ditto

WORKSPACE
```diff
@@ -217,7 +217,6 @@ maven_install(
     ],
     fetch_sources = True,
     repositories = [
-        "https://jcenter.bintray.com",
```
steveniemitz, line 220: jcenter was shut down awhile ago, and all the files return a 403 now.  I removed this because it was causing bazel to hang resolving the 3rdparty deps.

server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepServer.java
steveniemitz: +1, although you can compile with a language target of java 11 but still produce bytecode for java 8 (which seems to be what was happening in my case).

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/TargetKindResolver.java
steveniemitz: is this resolved? no action then?

lukaszwawrzyk: Maybe let's revert and have both NO_IDE and LIBRARY for example on a target. I can imagine that we will be for example implementing import depth feature and traverse the tree of deps, but only for library, binary and test targets and we could skip some transitive dep because of this.

steveniemitz: oh wow I didn't actually realize that I changed the logic here.  That was unintentional

server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspAspectsManager.java
abrams27, line 34: i like colors as well

server/src/main/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticBspMapper.kt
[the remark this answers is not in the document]
steveniemitz: alright I refactored this a little more.

steveniemitz:
I opened https://github.com/JetBrains/intellij-scala/pull/618 to fix one of the big issues w/ the module merger, it ended up creating VERY long target names that couldn't be persisted.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/TargetKindResolver.java
```diff
@@ -7,22 +7,29 @@
...
-          "test", Tag.TEST);
+          "library", LIBRARY,
+          "binary", HashSet.of(Tag.APPLICATION),
+          "test", HashSet.of(Tag.TEST));
+
+  private static final Set<Tag> NO_IDE = HashSet.of(Tag.NO_IDE);
 
   public Set<Tag> resolveTags(TargetInfo targetInfo) {
+    if (targetInfo.getKind().equals("resources_union")) {
+      return LIBRARY;
```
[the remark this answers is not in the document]
steveniemitz: no this actually needs to be a "normal" target.  I'm pretty sure this is working around a twitter specific bazel quirk, but this is needed to maintain the correct dependency graph, otherwise things that depend on resources won't correctly import.

server/src/test/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticsServiceTest.kt
```diff
@@ -398,7 +398,7 @@ class DiagnosticsServiceTest {
                    |FAILED: Build did NOT complete successfully""".trimMargin()
 
     // when
-    val diagnostics = extractDiagnostics(output)
+    val diagnostics = extractDiagnostics(output, "")
```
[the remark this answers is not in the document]
steveniemitz: sorry what's the question here?

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/TargetKindResolver.java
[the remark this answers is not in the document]
abrams27: i meant that this rule should support the `no-ide` tag as well as any other, but if there is no such use case then fine

server/src/test/java/org/jetbrains/bsp/bazel/server/diagnostics/DiagnosticsServiceTest.kt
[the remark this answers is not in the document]
abrams27: u updated tests before changing diagnostics parser logic - for java it was impossible to obtain target name from the output, so the name was blank, but now we are propagating the name, so u can set some mock name here.

(sorry for nitpicking - we can leave it as it is)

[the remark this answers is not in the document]
steveniemitz: ooo I see what you mean, yeah lemme do that.

steveniemitz:
oops, it looks like one of those suggestions I applied from the UI added an extra curly brace, I fixed it, can you rerun the tests?

PR #246, opened 2022-05-18
[feature] add the ability to use bloop as the bsp server

steveniemitz:
This needs a little more cleanup, but is in a spot where it's ready for review (and general use!)

I've been using it for a daily-driver now instead of fastpass for a week or so now and it's at least at parity (and better in most cases imo).

Most of the logic is in BloopExporter, most translation from the bazel-bsp model to bloop is pretty straight forward, but there's a couple places where we diverge slightly:

- Dependencies of included targets that themselves are included need to be remapped so we rely on the compile artifacts from bloop, not from bazel (otherwise changes made in the IDE don't have any affect).  This is handled in `fixupClassPath`
- We try to "reglob" sources to pass to bloop.  This allows uses to add new files to a directory without needing to re-import the project again, and also significantly reduces the size of the bloop project files.


server/src/main/java/org/jetbrains/bsp/bazel/server/BloopExporter.java
```diff
@@ -0,0 +1,549 @@
...
+              scala.Some.apply(classPath),
+              scala.Option.empty());
+        });
+  }
+
+  private Config.Test bloopTestFramework() {
+    var framework =
+        new Config.TestFramework(
+            CollectionConverters.asScala(
+                    Collections.singleton("munit.internal.junitinterface.PantsFramework"))
```
Duhemm, line 255: Are you sure this test interface will be on the classpath? It is normally provided by `org.scalameta:junit-interface:1.0.0-M3`. Some test classes will not be detected as such if the test interface is not on the (runtime) classpath.

server/src/main/java/org/jetbrains/bsp/bazel/server/BloopExporter.java
```diff
@@ -0,0 +1,549 @@
...
+  private Map<URI, URI> localArtifacts;
+
+  public BloopExporter(
+      BspInfo bspInfo, Path workspaceRoot, WorkspaceContextProvider workspaceContextProvider) {
+    this.bspInfo = bspInfo;
+    this.workspaceContextProvider = workspaceContextProvider;
+    this.workspaceRoot = workspaceRoot;
+  }
+
+  private void initializeClient(
```
abrams27, line 76: methid should be under the calling method

server/src/main/java/org/jetbrains/bsp/bazel/server/BloopExporter.java
```diff
@@ -0,0 +1,549 @@
...
+                      CollectionConverters.asScala(Collections.singleton(artifact)).toList());
+                });
+    return new Config.Resolution(CollectionConverters.asScala(resolutionModules).toList());
+  }
+
+  private <T> scala.collection.immutable.List<T> emptyList() {
+    return List$.MODULE$.empty();
+  }
+
+  private Option<scala.collection.immutable.List<Config.SourcesGlobs>> reGlob(
```
abrams27, line 299: plus it's a huge method which reallllllyyy needs tests (and which needs to be smaller and in a separate class)

server/src/main/java/org/jetbrains/bsp/bazel/server/BloopExporter.java
```diff
@@ -0,0 +1,549 @@
...
+                .toList());
+    return new Config.Test(
+        CollectionConverters.asScala(Collections.singleton(framework)).toList(),
+        new Config.TestOptions(List$.MODULE$.empty(), List$.MODULE$.empty()));
+  }
+
+  private scala.collection.immutable.List<String> convertTags(io.vavr.collection.Set<Tag> tags) {
+    Set<String> ret;
+    if (tags.contains(Tag.TEST)) {
+      ret = HashSet.of("test");
```
abrams27, line 265: why just not return here?

abrams27 (changes requested):
thanks a lot for that <3 I'll try to test it on my own later

formatting fails

left some comments + code itself needs some changes (especially `BloopExporter`) but I can do it later in the following PR if u don't have time for that

server/src/main/java/org/jetbrains/bsp/bazel/server/BloopExporter.java
steveniemitz: +100 on tests!  Once things stabilize I'll definitely write a bunch of tests for it.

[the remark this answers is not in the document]
steveniemitz: yeah, I didn't really want to rely on any external tools (particularly anything twitter specific), nor shell out to something else again.  This really doesn't have to be exact anyways given limitations in IJ (you can't have two modules share the same directory).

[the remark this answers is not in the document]
abrams27: > IJ (you can't have two modules share the same directory).

we are working on that!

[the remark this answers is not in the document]
abrams27: i'm in favour of 're globing' sources (less data to transfer) but in a proper way - without implicit merging or similar things

steveniemitz:
> formatting fails

its really weird, my google-java-format plugin won't wrap that string.  I'll do it manually, just haven't gotten around to it yet.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.java
```diff
@@ -101,6 +101,10 @@ private Seq<Label> resolveDirectDependencies(TargetInfo target) {
   }
 
   private Set<Language> inferLanguages(TargetInfo target) {
+    if (target.getKind().equals("scala_binary")) {
```
abrams27, line 104: is it an edge case for `scala_binary` without sources? the mechanizm below should work for any case with sources

abrams27: cool, maybe it means that we should adjust our mechanism in general - it always should return the language based on the target kind + try to detect more from sources - just a thing to consider later

steveniemitz:
sorry, I've been slacking on this review.  In the mean time I've been using this for the last couple weeks as my daily driver and its working great!  Besides adding unit tests, are there any other major blockers you see to merging this?

abrams27:
> sorry, I've been slacking on this review. In the mean time I've been using this for the last couple weeks as my daily driver and its working great! Besides adding unit tests, are there any other major blockers you see to merging this?

well, if u really want it merged, i can take care of tests and similar things later - u just need to rebase it

steveniemitz:
> well, if u really want it merged, i can take care of tests and similar things later - u just need to rebase it

oh no that's ok, I'm perfectly happy to do the work to get it up to speed.  I also noticed you left a lot of comments a couple weeks ago I missed, let me go over those as well.

install/src/main/java/org/jetbrains/bsp/bazel/install/BloopBspConnectionDetailsCreator.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.jetbrains.bsp.bazel.commons.Constants
+import org.jetbrains.bsp.bazel.installationcontext.InstallationContext
+import java.net.URL
+import java.nio.file.Files
+import java.nio.file.Path
+
+class BloopBspConnectionDetailsCreator(bazelBspPath: Path) {
+    private val coursierDestination = bazelBspPath.resolve("cs")
+
+    private fun downloadCoursier(): Try<Void> {
```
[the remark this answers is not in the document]
steveniemitz: hm, I'm not following what you mean with a `when` expression here, I'm also not super familiar with kotlin. :P  Wouldn't each test of the when need to be using the same input?

Also I do think it's useful to install coursier here, at least in this case most people aren't launching with coursier, and I'm not sure how many people will have it installed by default in the wild.

steveniemitz:
@abrams27 did a big code cleanup and added some tests, let me know what you think!

steveniemitz:
ping @abrams27 :)

install/src/main/java/org/jetbrains/bsp/bazel/install/BloopEnvironmentCreator.kt
```diff
@@ -0,0 +1,71 @@
...
+    private fun createBloopConfig(path: Path): Try<Void> =
+        refreshProjectArgs().flatMap {
+            val settings = WorkspaceSettings(it)
+            val bloopSettingsJsonPath = path.resolve(Constants.BLOOP_SETTINGS_JSON_FILE_NAME)
+            writeJsonToFile(bloopSettingsJsonPath, settings)
+        }
+
+
+    private fun refreshProjectArgs(): Try<List<String>> {
+        val pwd = Paths.get("").toAbsolutePath()
```
abrams27, line 58: so the question why do we want to know this directory? it's a directory where the installer was called - it can be called anywhere

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporterInitializer.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jetbrains.bsp.bazel.workspacecontext.DefaultWorkspaceContextProvider;
+
+public class BloopExporterInitializer {
+  public static void main(String[] args) {
+    Option<Path> projectViewPath;
+    Path bspProjectRoot = Paths.get("").toAbsolutePath().getParent();
+    Path projectRoot;
+
+    if (args.length == 1) {
+      projectRoot = Paths.get(args[0]);
```
abrams27, line 18: it's not a project root - it's a pwd of the installer

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporter.java
```diff
@@ -0,0 +1,166 @@
...
+    initializeClient(serverContainer, client);
+
+    Project project = projectProvider.refreshAndGet();
+
+    var failedTargets = client.getFailedTargets();
+    var failedTransitiveTargets =
+        failedTargets.removeAll(
+            project.modules().map(m -> new BuildTargetIdentifier(m.label().getValue())));
+
+    if (failedTransitiveTargets.nonEmpty()) {
```
abrams27, line 61: does it mean that if anything in transitive deps fails an entire import fails? we are trying to collect as much data as we can, even some targets are not correct

server/src/main/java/org/jetbrains/bsp/bazel/server/common/BUILD
```diff
@@ -0,0 +1,38 @@
...
+        "@maven//:com_google_guava_guava",
+        "@maven//:io_vavr_vavr",
+        "@maven//:org_apache_logging_log4j_log4j_api",
+        "@maven//:org_apache_logging_log4j_log4j_core",
+        "@maven//:org_eclipse_lsp4j_org_eclipse_lsp4j_jsonrpc",
+        "@maven//:org_scala_lang_scala_library",
+    ],
+)
+
+java_binary(
```
abrams27, line 33: should it be here?

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BspProjectExporter.java
```diff
@@ -0,0 +1,94 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.jetbrains.bsp.bazel.server.sync.languages.LanguageData;
+import org.jetbrains.bsp.bazel.server.sync.languages.java.JavaModule;
+import org.jetbrains.bsp.bazel.server.sync.languages.scala.ScalaModule;
+import org.jetbrains.bsp.bazel.server.sync.model.Module;
+import org.jetbrains.bsp.bazel.server.sync.model.Project;
+
+class BspProjectExporter {
+  private static final Set<Path> IGNORED_SOURCES =
```
abrams27, line 18: i dont like coupling the implementation with the twitter codebase - we need to come up with a better mechanism

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/ReGlobber.java
```diff
@@ -0,0 +1,71 @@
...
+import java.nio.file.Paths;
+import org.jetbrains.bsp.bazel.server.sync.model.SourceSet;
+import scala.collection.immutable.List;
+
+class ReGlobber {
+  public static ReGlobbed reGlob(URI baseDirectory, SourceSet sourceSet) {
+    scala.collection.immutable.List<Path> sources;
+    var sourceGlobs = reGlobImpl(baseDirectory, sourceSet.sources());
+    if (sourceGlobs.isDefined()) {
+      sources = emptyList();
```
abrams27, line 20: why just not return `ReGlobbed` here instead of c-like reassignments?

server/src/test/java/org/jetbrains/bsp/bazel/server/bloop/ReGlobberTest.kt
```diff
@@ -0,0 +1,60 @@
...
+import io.kotest.matchers.shouldBe
+import org.jetbrains.bsp.bazel.server.sync.model.SourceSet
+import org.junit.jupiter.api.Test
+import java.nio.file.Paths
+import io.vavr.collection.HashSet
+import org.jetbrains.bsp.bazel.server.bloop.ScalaInterop.emptyList
+import org.jetbrains.bsp.bazel.server.bloop.ScalaInterop.toList
+import scala.jdk.javaapi.CollectionConverters
+
+class ReGlobberTest {
```
abrams27, line 14: how about the case when user specifies files by name? of example only one in the directory?

abrams27 (changes requested):
sorry, i was quite busy

i left some comments / questions

also it'd be nice to mention in readme how to use it with bloop
+ update changelo

also, could u please describe how this integration works? what is the flow? i'd like to try it on my own, but im not sure exactly how it's suppose to work

install/src/main/java/org/jetbrains/bsp/bazel/install/BloopEnvironmentCreator.kt
steveniemitz: the way we've been doing it right now is calling the installer from the root of the repository, so `pwd` == repository root.

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporterInitializer.java
steveniemitz: in this case its used as the project root.

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporter.java
steveniemitz: yeah, modules inside the project can fail, but ones outside will fail the import.  I this this is the best compromise:
- stuff being actively developed breaking will still allow you to import
- things outside of the project breaking will fail it, rather than giving bizarre errors at runtime or compile time.

server/src/test/java/org/jetbrains/bsp/bazel/server/bloop/ReGlobberTest.kt
steveniemitz: right now it'll still reglob anything, even if there the sources don't represent the entire directory.

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BspProjectExporter.java
abrams27: yea, we'll definitely come up with something better in the future

server/src/test/java/org/jetbrains/bsp/bazel/server/bloop/ReGlobberTest.kt
abrams27: is it an expected behavior / what is the reason?

install/src/main/java/org/jetbrains/bsp/bazel/install/BloopEnvironmentCreator.kt
steveniemitz: oh I think there's a difference here.  The "workspace root" is where bazel-bsp writes its files (as well as bloop in this case).  This (for our use case at least) is outside of the "repository root".  We could add another flag for repository root to make this explicit, using the current working directory is a little janky I agree.

server/src/test/java/org/jetbrains/bsp/bazel/server/bloop/ReGlobberTest.kt
abrams27: > IJ doesn't support more complicated globs

we'll fix it soon

install/src/main/java/org/jetbrains/bsp/bazel/install/BloopEnvironmentCreator.kt
abrams27: ohh i see, it seems like a bazel-bsp limitation as well - https://youtrack.jetbrains.com/issue/BAZEL-108

steveniemitz: Cool, yeah.  With these changes I think it'd be pretty easy now to make the normal bazel-bsp support it too.

abrams27: yup, but i think we can do it in a separate PR

install/src/main/java/org/jetbrains/bsp/bazel/install/cli/CliOptionsProvider.kt
```diff
@@ -136,6 +136,18 @@ class CliOptionsProvider(private val args: Array<String>) {
...
+        val useBloopOption = Option.builder(USE_BLOOP_SHORT_OPT)
+            .longOpt("use_bloop")
+            .build()
+        cliParserOptions.addOption(useBloopOption)
+
+        val projectNameOption = Option.builder(PROJECT_NAME_SHORT_OPT)
+            .longOpt("project_name")
+            .hasArg()
+            .argName("name")
+            .build()
```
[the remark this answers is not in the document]
steveniemitz: This isn't even actually used, oops, removed.

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelRunner.kt
```diff
@@ -37,9 +37,7 @@ class BazelRunner private constructor(
   fun commandBuilder(): BazelRunnerCommandBuilder = BazelRunnerCommandBuilder(this)
 
   fun runBazelCommandBes(command: String, flags: List<String>, arguments: List<String>): BazelProcess {
-    fun besFlags() = listOf(
-        "--bes_backend=grpc://localhost:${besBackendPort!!}",
-        "--build_event_publish_all_actions")
+    fun besFlags() = listOf("--bes_backend=grpc://localhost:${besBackendPort!!}")
```
abrams27, line 40: dont we need this flag?

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BUILD
```diff
@@ -0,0 +1,39 @@
...
+        "@maven//:com_google_guava_guava",
+        "@maven//:io_vavr_vavr",
+        "@maven//:org_apache_logging_log4j_log4j_api",
+        "@maven//:org_apache_logging_log4j_log4j_core",
+        "@maven//:org_eclipse_lsp4j_org_eclipse_lsp4j_jsonrpc",
+        "@maven//:org_scala_lang_scala_library",
+    ],
+)
+
+java_binary(
```
abrams27, line 34: shoukd it be here as well?

abrams27 (commented):
I'll try it out tomorrow, and then we can merge it

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelRunner.kt
steveniemitz: we don't as far as I can tell.  According to the [bazel BEP reference](https://docs.bazel.build/versions/main/bep-glossary.html#actionexecuted) the flag only controls sending successful `ActionExecuted` events, but we only care about failed `ActionExecuted`, which are sent either way. https://github.com/JetBrains/bazel-bsp/blob/master/server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepServer.java#L137

abrams27: ohhhh, cool, my bad

abrams27:
I installed bloop (using brew), then `.install.sh --use_bloop` in the project, the import was successful but im getting red code : ( - the dependencies are not resolved properly. Any ideas?

steveniemitz:
> I installed bloop (using brew), then `./install.sh --use_bloop` in the project, the import was successful but im getting red code : ( - the dependencies are not resolved properly. Any ideas?

You shouldn't have to install bloop, the installer downloads it for you.  You might need to pass a target spec (`-t`)? Try `-t //...` for everything.  I've never tried running it without both `-t` and `-d`

abrams27:
hmm is it possible to see what bloop sends to the client? deps are only mentioned it `//target dependencies` - the client usually puts them there if they are not provided with javaopts

steveniemitz:
> bazel-bsp xd

heh, yeah it seems like the kotlin dependencies are really throwing bloop off.  //commons compiles fine:
```
bloop compile commons/src/main/java/org/jetbrains/bsp/bazel/commons:commons
```

but it does still show up as missing deps in IJ.

steveniemitz:
think we can merge this in today?

abrams27:
I checked it with another project, seems working, soo okeeeyyy lets goo

abrams27 (approved):
I checked it with another project, seems working, soo okeeeyyy lets goo

PR #260, opened 2022-06-17
[change] JvmEnvironmentItem is set using `env` from target

basakkapusuzoglu:
env in JvmEnvironmentItem should be set using env and env_inherit

---

bazel run //e2e:BazelBspSampleRepoTest


install/src/main/resources/aspects.bzl
```diff
@@ -482,6 +482,9 @@ def _bsp_target_info_aspect_impl(target, ctx):
     java_toolchain_info, java_toolchain_info_exported = extract_java_toolchain(target, ctx, dep_targets)
     java_runtime_info, java_runtime_info_exported = extract_java_runtime(target, ctx, dep_targets)
 
+    print(rule_attrs.env)
```
agluszak, line 485: don't forget to remove that before submitting the final version :)

install/src/main/resources/aspects.bzl
```diff
@@ -494,6 +497,8 @@ def _bsp_target_info_aspect_impl(target, ctx):
         java_target_info = java_target_info,
         java_toolchain_info = java_toolchain_info,
         java_runtime_info = java_runtime_info,
+        env = rule_attrs.env,
```
agluszak, line 500: I think it can fail if there is no `env` attribute, please take a look at https://github.com/JetBrains/bazel-bsp/blob/f77d734f631b8d9875c0a5f842504a236785be52/install/src/main/resources/aspects.bzl#L587=

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.java
```diff
@@ -139,4 +141,12 @@ private Map<URI, Label> buildReverseSourceMapping(Seq<Module> modules) {
         });
     return HashMap.ofAll(output);
   }
+
+  private Map<String, String> environmentItem(TargetInfo target) {
+    if (target.containsEnvInherit("80")) {
```
agluszak, line 146: what is "80"?

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaLanguagePlugin.java
```diff
@@ -117,7 +116,7 @@ public JvmEnvironmentItem toJvmEnvironmentItem(Module module, JavaModule javaMod
         javaModule.runtimeClasspath().map(URI::toString).asJava(),
         javaModule.jvmOps().asJava(),
         bazelInfo.getWorkspaceRoot().toString(),
-        Map.of());
+        module.getEnvironment());
     // FIXME: figure out what we should pass here, because passing the environment
```
agluszak, line 120: you can remove that FIXME

install/src/main/resources/aspects.bzl
agluszak:
```
env = dict()
if hasattr(rule_attrs, "env") {
  env = rule_attrs.env
}
```

install/src/main/resources/aspects.bzl
```diff
@@ -494,6 +497,8 @@ def _bsp_target_info_aspect_impl(target, ctx):
...
+        env_inherit = rule_attrs.env_inherit,
```
agluszak, line 501: same as above

agluszak (changes requested):

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.java
basakkapusuzoglu: it is the number I assigned to env_inherit in bsp_target_info.proto

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/proto/bsp_target_info.proto
```diff
@@ -64,6 +64,8 @@ message TargetInfo {
   repeated Dependency dependencies = 40;
   repeated FileLocation sources = 50;
   repeated FileLocation resources = 60;
+  map <string, string> env = 70;
+  map <string , string> env_inherit = 80;
```
agluszak, line 68:
```suggestion
  map <string, string> env_inherit = 80;
```

install/src/main/resources/aspects.bzl
abrams27: u can just use `getattr`

abrams27: https://docs.bazel.build/versions/main/skylark/lib/globals.html#getattr

abrams27 (changes requested):
it could be tested in e2e tests - just add another dir in sample repo and set some envs there

e2e/test-resources/sample-repo/environement_variables/BUILD
```diff
@@ -0,0 +1,16 @@
...
+    name = "java_library",
+    srcs = ["JavaEnv.java"],
+    env = {"foo1": "foo2"},
+    visibility = ["//visibility:public"],
+)
+
+java_test(
+    name = "java_test",
+    srcs = ["JavaTest.java"],
+    env_inherit = ["foo3"],
```
abrams27, line 13: u can set `env` here as well

e2e/test-resources/sample-repo/environement_variables/BUILD
```diff
@@ -0,0 +1,16 @@
+load("@rules_java//java:defs.bzl", "java_library", "java_test")
+
+java_library(
+    name = "java_library",
+    srcs = ["JavaEnv.java"],
+    env = {"foo1": "foo2"},
```
abrams27, line 6: it's available in binary / test not in libabry


plus u can set these variables in tests (for env_iherit as well, otherwise inherited envs would be empty) in e2e bash scripts https://github.com/JetBrains/bazel-bsp/blob/master/e2e/runTest.sh <- u can add an optional field with envs set for the execution

e2e/src/main/java/org/jetbrains/bsp/bazel/BazelBspSampleRepoTest.kt
```diff
@@ -1025,6 +1053,30 @@ object BazelBspSampleRepoTest : BazelBspTestBaseScenario() {
...
+            listOf("library"),
+            listOf("java"),
+            emptyList(),
+            BuildTargetCapabilities(true, true, false, false)
+        )
+        environmentVariablesJavaTest.displayName = "//environment_variables:java_test"
+        environmentVariablesJavaTest.baseDirectory = "file://\$WORKSPACE/environment_variables/"
+        environmentVariablesJavaTest.dataKind = "jvm"
+        environmentVariablesJavaTest.data = jvmBuildTarget
+
```
abrams27, line 1079: u need to update env options calls as well

abrams27 (changes requested):

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.java
```diff
@@ -145,4 +148,12 @@ private Map<URI, Label> buildReverseSourceMapping(Seq<Module> modules) {
         });
     return HashMap.ofAll(output);
   }
+
+  private Map<String, String> environmentItem(TargetInfo target) {
+    var output =
```
abrams27, line 153: u can extract it to a method

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.java
```diff
@@ -145,4 +148,12 @@ private Map<URI, Label> buildReverseSourceMapping(Seq<Module> modules) {
...
+        target.getEnvInheritList().stream().collect(Collectors.toMap(x -> x, System::getenv));
+    var result = Stream.of(output)
```
abrams27, line 155: same and merge results in this method

e2e/src/main/java/org/jetbrains/bsp/bazel/BazelBspSampleRepoTest.kt
```diff
@@ -655,6 +683,24 @@ object BazelBspSampleRepoTest : BazelBspTestBaseScenario() {
...
                     mapOf()
+                ),
+                JvmEnvironmentItem(
+                    BuildTargetIdentifier("//environment_variables:java_binary"),
+                    listOf(
+                        "file://\$BAZEL_CACHE/bazel-out/k8-fastbuild/bin/environment_variables/java_binary.jar"
+                    ),
+                    emptyList(),
+                    "\$WORKSPACE",
+                    mapOf()
```
abrams27, line 694: i think that this tests has sense only if this map wont be empty

abrams27 (changes requested):

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.java
```diff
@@ -145,4 +148,15 @@ private Map<URI, Label> buildReverseSourceMapping(Seq<Module> modules) {
         });
     return HashMap.ofAll(output);
   }
+
+  private Map<String, String> environmentItem(TargetInfo target) {
+    var output = targetInfoToMapOfStrings(target);
+    var targetEnv = target.getEnvMap();
+    return output.merge((Map<? extends String, ? extends String>) targetEnv);
```
abrams27, line 155: it's a wrong casting - in general u want to avoid `(type)` casting, U can just wrap it with `HashMap.ofAll(....)` - it's one of the reasons why tests are failing

e2e/runTest.sh
```diff
@@ -15,6 +15,8 @@ TEST_TARGET="$1"
 # second argument (optional) of the script should be a path to the directory with tested project (relative to the project root)
 TEST_PROJECT_PATH="$2"
 
+export env = "foo1 : foo2"
```
abrams27, line 18: in bash the syntax is
```
export NAME=VALUE
```

also, we'd like to make it a bit more generic - these envs could be read from the arguments, and then we can set them only for the specific test in BUIOLD file

also it's possible to set them only for the single bash call:
```
NAME=VALUE ./bash_srcipt.sh
```

it's cleaner

abrams27 (changes requested):

e2e/test-resources/sample-repo/environement_variables/BUILD
```diff
@@ -0,0 +1,18 @@
+load("@rules_java//java:defs.bzl", "java_library", "java_test")
+
+java_library(
+    name = "java_binary",
+    srcs = ["JavaEnv.java"],
+    env = {"foo1": "foo2"},
```
abrams27, line 6:
```suggestion
env = {"foo1": "val1", "foo2" : "val2" },
```

e2e/test-resources/sample-repo/environement_variables/BUILD
```diff
@@ -0,0 +1,18 @@
...
+    main_class = "environment_variables.JavaEnv",
+    visibility = ["//visibility:public"],
+)
+
+java_test(
+    name = "java_test",
+    srcs = ["JavaTest.java"],
+    env = {"foo3": "foo4"},
```
abrams27, line 14:
```suggestion
env = {"foo1": "val1", "foo2" : "val2" },
```

e2e/test-resources/sample-repo/environement_variables/BUILD
```diff
@@ -0,0 +1,18 @@
...
+    env_inherit = ["foo5"],
```
abrams27, line 15: "foo3", "foo4"

e2e/runTest.sh
abrams27:
```
foo1=invalid_val1 foo2=invalid_val2 foo3=val3 foo4=val4 bazel run "$TEST_TARGET"
```

abrams27 (changes requested):

abrams27 (changes requested):

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.kt
```diff
@@ -129,4 +135,15 @@ class BazelProjectMapper(
         with(module) {
             (sourceSet.sources + resources).map { Pair(it, label) }
         }
+
+    private fun environmentItem(target: TargetInfo): Map<String, String> {
+        val inheritedEnvs = collectInheritedEnvs(target)
+        val targetEnv = target.envMap
+        return inheritedEnvs.toMap(targetEnv)
```
abrams27, line 142:
```suggestion
        return inheritedEnvs + targetEnv
```

its the reason why tests are failing - toMap puts values into the destination, and in this case it fails

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.kt
```diff
@@ -129,4 +135,15 @@ class BazelProjectMapper(
...
+    }
+
+    private fun collectInheritedEnvs(targetInfo: TargetInfo): Map<String, String> {
```
abrams27, line 145:
```suggestion
    private fun collectInheritedEnvs(targetInfo: TargetInfo): Map<String, String> =
```

abrams27 (changes requested):

abrams27 (approved):
please remove the empty `Module.java` file

PR #264, opened 2022-07-05
[feature] add bazelWorkspaceRootDir flag

basakkapusuzoglu:


install/src/main/java/org/jetbrains/bsp/bazel/install/BloopEnvironmentCreator.kt
```diff
@@ -53,15 +54,14 @@ class BloopEnvironmentCreator(
...
-        val pwd = Paths.get("").toAbsolutePath()
         return launcherArgumentCreator.classpathArgv().map {
             listOfNotNull(
                 launcherArgumentCreator.javaBinaryArgv(),
                 Constants.CLASSPATH_FLAG,
                 it,
                 launcherArgumentCreator.debuggerConnectionArgv(),
                 Constants.BLOOP_BOOTSTRAP_CLASS_NAME,
-                pwd.toString(),
+                launcherArgumentCreator.bazelWorkspaceRootDir(),
```
steveniemitz, line 64: I think this changes the semantics of how this used to work.  Before the change it'd always use the current working directory, after the change, if `bazelWorkspaceRootDir` is unset it'll use `workspaceRootDir` first (if set).

I'm actually +1 to the change still, just wanted to make sure I'm understanding what's changing.  We should update the README as well.

abrams27: yea we want to have the following order:
1. bazel workspace root dir flag (`-w`)
2. workspace root dir (`-d`)
3. pwd

I think that we need such order because the "normal" use case is to install server with `-d` inside bazel workspace, and we'd like to keep this behavior without forcing users to use one more flag with exactly the same path

install/src/main/java/org/jetbrains/bsp/bazel/install/Install.kt
```diff
@@ -55,6 +56,12 @@ object Install {
         return environmentCreator.create()
     }
 
+    private fun createEnvironmentOutside(details: BspConnectionDetails, cliOptions: CliOptions): Try<Void> {
```
abrams27, line 59: what is the purpose of this method?

executioncontext/installationcontext/src/test/java/org/jetbrains/bsp/bazel/installationcontext/InstallationContextConstructorTest.kt
```diff
@@ -20,7 +20,7 @@ class InstallationContextConstructorTest {
     fun beforeEach() {
         // given
         this.installationContextConstructor =
-            InstallationContextConstructor(Path("/path/to/projectview.bazelproject"))
+            InstallationContextConstructor(Path("/path/to/projectview.bazelproject"), Path("/path/to/bazel.workspacerootdir"))
```
abrams27, line 23: its a dir
```suggestion
            InstallationContextConstructor(Path("/path/to/projectview.bazelproject"), Path("/path/to/bazel/workspace"))
```

install/src/main/java/org/jetbrains/bsp/bazel/install/Install.kt
basakkapusuzoglu: it is creating an environment for provided Bazel workspace root dir

abrams27: I believe that the existing mechanism could be still used

abrams27:
done in https://github.com/JetBrains/bazel-bsp/pull/276

PR #268, opened 2022-07-12
[maintenance] serialization cleanup

abrams27:
- i removed jackson annotations
- i removed custom `toString` implementations, since we are using data classes
- so also, i removed part of `Format` related to `toString` impls


server/src/test/java/org/jetbrains/bsp/bazel/server/bloop/BspModuleExporterTest.kt
```diff
@@ -78,8 +78,8 @@ class BspModuleExporterTest {
             baseDirectory.toScalaList().toOption(),
             emptyScalaList(),
             scalaListOf(baseDirectory.resolve("cp1")),
-            bloopRoot.resolve("z_73CA8EDE4AC5"),
```
lukaszwawrzyk, line 81: how could it break?

abrams27: toString for `Label` has changed and the hash in bloop is different now

abrams27: imo no - i can use `Naming.compilerOutputNameFor` there

lukaszwawrzyk (approved):

PR #270, opened 2022-07-13
[maintenance] bsp client logger is now an object :o

abrams27:
well, im not sure about this change, but i like the idea that we dont need to pass around this logger


lukaszwawrzyk:
It is not that much passing around. I would keep it as a class. It allows to swap it in tests, or swap it with a simpler implementation that just logs to stdout (for bloop case). As we refactor bazel-bsp further, it may even lose "bsp" from it's name. The project sync part doesn't need to know about bsp at all for example, it should see some logger interface.

abrams27:
agree, ill close this pr - we will get back to it one day during further refactors

PR #271, opened 2022-07-13
Adding ideClasspath filtering based on root target dependencies

sellophane:


server/src/main/java/org/jetbrains/bsp/bazel/server/sync/dependencytree/DependencyTree.kt
```diff
@@ -69,11 +69,19 @@ class DependencyTree(
...
                     .filter(::isNotARootTarget)
                     .flatMap(::collectTransitiveDependenciesAndAddTarget).toSet()
 
+    fun transitiveRootTargetDependencies(targetId: String): Set<TargetInfo> =
+            idToTargetInfo[targetId]?.let(::getDependencies).orEmpty()
+                    .flatMap {
+                        if (!isNotARootTarget(it))
+                            idToTargetInfo[it]?.let(::setOf).orEmpty()
+                        else
+                            transitiveRootTargetDependencies(it)
```
abrams27, line 82: and if we want a recursion let's do it a tailrec

abrams27 (changes requested):

PR #273, opened 2022-07-15
[change] added cache of targets info in incremental syncs

mikkoziel:
I moved reading target info from ProjectResolver class to new ProjectFileParser class and added HashMap caching info about Target with LastModified timestamps for every file to optimise parsing file that hadn't had any changes between incremental Syncs. Thank to that I was able to reduce the time of next sync (for example from ca. 20s in first import to ca. 300ms to next import without changes in files)

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/common/ServerContainer.kt
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/common/ServerContainer.kt
@@ -11,13 +11,7 @@ import org.jetbrains.bsp.bazel.server.bsp.info.BspInfo
 import org.jetbrains.bsp.bazel.server.bsp.managers.BazelBspAspectsManager
 import org.jetbrains.bsp.bazel.server.bsp.managers.BazelBspCompilationManager
 import org.jetbrains.bsp.bazel.server.bsp.utils.InternalAspectsResolver
-import org.jetbrains.bsp.bazel.server.sync.BazelPathsResolver
-import org.jetbrains.bsp.bazel.server.sync.BazelProjectMapper
-import org.jetbrains.bsp.bazel.server.sync.FileProjectStorage
-import org.jetbrains.bsp.bazel.server.sync.ProjectProvider
-import org.jetbrains.bsp.bazel.server.sync.ProjectResolver
-import org.jetbrains.bsp.bazel.server.sync.ProjectStorage
-import org.jetbrains.bsp.bazel.server.sync.TargetKindResolver
+import org.jetbrains.bsp.bazel.server.sync.*
 import org.jetbrains.bsp.bazel.server.sync.languages.LanguagePluginsService
 import org.jetbrains.bsp.bazel.server.sync.languages.cpp.CppLanguagePlugin
 import org.jetbrains.bsp.bazel.server.sync.languages.java.JavaLanguagePlugin
@@ -72,11 +66,13 @@ class ServerContainer internal constructor(
             val targetKindResolver = TargetKindResolver()
             val bazelProjectMapper =
                 BazelProjectMapper(languagePluginsService, bazelPathsResolver, targetKindResolver)
+            val targetInfoReader = TargetInfoReader()
             val projectResolver = ProjectResolver(
                 bazelBspAspectsManager,
                 workspaceContextProvider,
                 bazelProjectMapper,
-                bspClientLogger
+                bspClientLogger,
+                targetInfoReader
             )
             val finalProjectStorage = projectStorage ?: FileProjectStorage(bspInfo, bspClientLogger)
             val projectProvider = ProjectProvider(projectResolver, finalProjectStorage)
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
@@ -1,11 +1,5 @@
 package org.jetbrains.bsp.bazel.server.sync
 
-import com.google.protobuf.TextFormat
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.flow.asFlow
-import kotlinx.coroutines.flow.map
-import kotlinx.coroutines.flow.toList
-import kotlinx.coroutines.runBlocking
 import org.jetbrains.bsp.bazel.info.BspTargetInfo.TargetInfo
 import org.jetbrains.bsp.bazel.logger.BspClientLogger
 import org.jetbrains.bsp.bazel.server.bep.BepOutput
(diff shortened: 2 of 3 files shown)


server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
```diff
@@ -0,0 +1,50 @@
...
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.nio.file.attribute.FileTime
+import kotlin.io.path.getLastModifiedTime
+
+class ProjectFileParser {
+
+    data class TimeStampInfo(val timestamp: FileTime, val bspTargetInfo: BspTargetInfo.TargetInfo)
+
+    private val timestampsMap: HashMap<URI, TimeStampInfo> = hashMapOf()
```
lukaszwawrzyk, line 21: maybe name it `cache` to be clear

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
```diff
@@ -0,0 +1,50 @@
...
+import java.net.URI
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.nio.file.attribute.FileTime
+import kotlin.io.path.getLastModifiedTime
+
+class ProjectFileParser {
+
+    data class TimeStampInfo(val timestamp: FileTime, val bspTargetInfo: BspTargetInfo.TargetInfo)
```
lukaszwawrzyk, line 19: Timestamp is part of the info, but it is info about TargetInfoFile. I would call it maybe `CachedTargetInfo`

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.coroutines.runBlocking
+import org.jetbrains.bsp.bazel.info.BspTargetInfo
+import java.net.URI
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.nio.file.attribute.FileTime
+import kotlin.io.path.getLastModifiedTime
+
+class ProjectFileParser {
```
lukaszwawrzyk, line 17: maybe TargetInfoReader. ProjectFile can be a confusing name, as we have 4 project models in bazell-bsp (from aspect, common, bsp and bloop).

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    private val timestampsMap: HashMap<URI, TimeStampInfo> = hashMapOf()
+
+    fun readTargetMapFromAspectOutputs(files: Set<URI>): Map<String, BspTargetInfo.TargetInfo> =
+        runBlocking(Dispatchers.IO) {
+            files.asFlow().map(::getTargetInfoForFileTimeStamp).toList()
+                .associateBy(BspTargetInfo.TargetInfo::getId)
+        }
+
+    private fun readTargetInfoFromFile(uri: URI): BspTargetInfo.TargetInfo {
```
lukaszwawrzyk, line 29: maybe import `BspTargetInfo.TargetInfo` to just write `TargetInfo` everywhere instead of `BspTargetInfo.TargetInfo`

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
```diff
@@ -0,0 +1,50 @@
...
+        val builder = BspTargetInfo.TargetInfo.newBuilder()
+        val parser = TextFormat.Parser.newBuilder().setAllowUnknownFields(true).build()
+        parser.merge(Files.newBufferedReader(Paths.get(uri), StandardCharsets.UTF_8), builder)
+        return builder.build()
+    }
+
+    private fun getTargetInfoForFileTimeStamp(uri: URI): BspTargetInfo.TargetInfo {
+        val lastModified = getFileTimeStamp(uri)
+        timestampsMap[uri]?.let { check ->
```
lukaszwawrzyk, line 38: I would change naming for example like this, for consistency:
```
cache[uri]?.let { cached -> if (cached.timestamp == currentTimestamp) return cached.targetInfo }
```
lets call timestamp timestamp, rather than lastModified, or let's call it lastModified everywhere.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+    private fun readTargetInfoFromFile(uri: URI): BspTargetInfo.TargetInfo {
+        val builder = BspTargetInfo.TargetInfo.newBuilder()
+        val parser = TextFormat.Parser.newBuilder().setAllowUnknownFields(true).build()
+        parser.merge(Files.newBufferedReader(Paths.get(uri), StandardCharsets.UTF_8), builder)
+        return builder.build()
+    }
+
+    private fun getTargetInfoForFileTimeStamp(uri: URI): BspTargetInfo.TargetInfo {
```
lukaszwawrzyk, line 36: it is not getting target info for file time stamp, it is getting target info from cache based on timestamp. I would just call it `getTargetInfo`, optionally `getOrReadTargetInfo`, `getTargetInfoCached`. Something like this.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
lukaszwawrzyk, line 21: it should be a ConcurrentHashMap as it is accessed from multiple threads

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
mikkoziel: Renamed

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
mikkoziel: Renamed

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectFileParser.kt
mikkoziel: Renamed

PR #279, opened 2022-07-26
[fix] Ensure AsyncOutputProcessor deterministically processes stdout

tpasternak:
1. When child process exited, `isRunning` flag was immediately set to
false, and the handlers were not called due to if(isRunning.get())
condition. It was very likely that a bunch of last lines were ignored
due to this.

2. In case stdout handlers execution took too long, the exception has
been silently ignored, despite very short timeout. So it was likely
that a line could be skipped even if the process was behaving correctly.

### DIFF
--- a/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/outputs/AsyncOutputProcessor.kt
+++ b/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/outputs/AsyncOutputProcessor.kt
@@ -7,7 +7,6 @@ import java.io.InputStreamReader
 import java.util.concurrent.Executors
 import java.util.concurrent.Future
 import java.util.concurrent.TimeUnit
-import java.util.concurrent.TimeoutException
 import java.util.concurrent.atomic.AtomicBoolean
 
 class AsyncOutputProcessor(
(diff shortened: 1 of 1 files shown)


lukaszwawrzyk (approved):

tpasternak:
@abrams27 so we have three successful integration test runs in a row https://github.com/JetBrains/bazel-bsp/actions/runs/2739862462

abrams27 (changes requested):
can u re-rebase?

tpasternak:
@abrams27 done

abrams27 (approved):
gooooood

steveniemitz:
I actually had done this on purose, there's a bug in java + linux around reading from a process's stdout/err that causes it to hang for ~45 seconds after the process exists.

https://bugs.openjdk.org/browse/JDK-8169565

Making it not wait was the only good way to avoid the deadlock (and a 45 second hang every time bazel was invoked).

steveniemitz:
Maybe we can pass in a "really wait" flag or something for bazel-info specifically?  Do other places actually need the output?  My assumption was that losing a couple lines of output from the bazel output itself wouldn't be a huge deal, but it makes sense you'd need it all for bazel info.

tpasternak:
Yeah, either a flag or a separate OutputProcessor, that could be used in places where the slowdown occurs and the exact output is not essential

tpasternak:
Personally, I would prefer a separate processor, as IMO most of the code would be different, depending on the flag

steveniemitz:
Yeah that seems reasonable to me, especially because `bazel info` won't be very rarely run, so a slow-down there isn't a huge deal.  My main goal was to get the standard "invoke bazel to do some building" loop to be faster.

abrams27:
also, we could run `bazel info workspace` instead of `bazel info` and the output would be shorter - just one line --> profit

lukaszwawrzyk:
But we are extracting exec-root as well

abrams27:
then `bazel info workspace execution_root`

PR #281, opened 2022-07-26
[maintenance] fix coursier command in README

steveniemitz:
coursier needs a `--` after the class to launch and arguments


abrams27 (approved):
thxx

PR #282, opened 2022-07-27
[fix] don't use Label.toString now that it's no longer overriden

steveniemitz:
https://github.com/JetBrains/bazel-bsp/commit/57c2c19c33e3055adeba9d043d9269435ddb83ac removed the override of Label.toString, which broke its representation in a couple places, this changes them to use `.value` instead.


steveniemitz:
cc @abrams27

abrams27 (approved):
ohh, sorry i forgot about bloop part, thanks

PR #284, opened 2022-08-08
[fix] workspace root is always used as a bazel directory

abrams27:
so now we are using workspace root always - even if user specifies a directory inside the workspace


bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelRunner.kt
```diff
@@ -15,6 +15,7 @@ class BazelRunner private constructor(
   companion object {
     private val LOGGER = LogManager.getLogger(BazelRunner::class.java)
 
+    // TODO: can we remove this logic?
```
lukaszwawrzyk, line 18: just remove it if it is unused now

lukaszwawrzyk (approved):

PR #261, opened 2022-06-27
Java to Kotlin rework for server/sync

sellophane:

PR #287, opened 2022-08-15
[fix] fix transitive target failure check in bloop export

steveniemitz:
#261 accidentally switched the predicate to test if any transitive targets failed, resulting in direct failures causing project imports to fail, and transitive ones to succeed.  This is the opposite of what should happen.

This changes it back, and adds a unit test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 ## [Unreleased]
 
+### Fixes 🛠️
+
+- Fix transitive target failure check in bloop export
+  | [#287](https://github.com/JetBrains/bazel-bsp/pull/287)
+
 ## [2.2.1] - 09.08.2022
 
 ### Fixes 🛠️
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporter.java
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporter.java
@@ -18,6 +18,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.jetbrains.bsp.bazel.server.bep.BepServer;
 import org.jetbrains.bsp.bazel.server.bsp.info.BspInfo;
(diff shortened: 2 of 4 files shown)


server/src/test/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporterTest.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.jupiter.api.fail
+
+class BloopExporterTest {
+    @Test
+    fun export_project_with_failed_direct_target_succeeds() {
+        val projectTargets = setOf(BuildTargetIdentifier("a"), BuildTargetIdentifier("b"))
+        val failedTargets = setOf(BuildTargetIdentifier("a"))
+
+        try {
+            BloopExporter.validateNoFailedExternalTargets(projectTargets, failedTargets)
```
steveniemitz, line 15: I wanted to use `assertThrows` etc here but using them caused the kotlin compiler to crash...

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporter.java
```diff
@@ -33,6 +34,19 @@ class BloopExporter {
   private final WorkspaceContextProvider workspaceContextProvider;
   private final Path workspaceRoot;
 
+  static void validateNoFailedExternalTargets(
+      Set<BuildTargetIdentifier> projectTargets, Set<BuildTargetIdentifier> failedTargets)
+      throws BazelExportFailedException {
+    var failedExternalTargets =
```
abrams27, line 40: it works for me, but it could be done with `.removeAll` - just calculate the difference and check is it empty or not

abrams27 (approved):

abrams27:
and please update the changelog

server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporter.java
steveniemitz: yeah, I do prefer treating things as immutable though, particularly mutating input parameters.

steveniemitz:
> and please update the changelog

done!

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
   | [#283](https://github.com/JetBrains/bazel-bsp/pull/283)
 - Don't use `Label.toString` now that it's no longer overriden.
   | [#282](https://github.com/JetBrains/bazel-bsp/pull/282)
+- Fix transitive target failure check in bloop export
```
abrams27, line 21: please add it in the `unreleased` section

abrams27 (changes requested):

CHANGELOG.md
steveniemitz: oops, done

abrams27 (approved):

PR #288, opened 2022-08-15
[change] add async and sync output processors

steveniemitz:
#279 re-introduced a bug on linux where the OutputProcessor can hang for up to 45 seconds waiting for the process (which has already existed) to exit.

This change adds a different type of OutputProcessor for commands that need to wait (bazel info, etc) to receive the full output, vs ones that don't (piping the build logs to the output handlers).

### DIFF
--- a/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.kt
+++ b/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.kt
@@ -15,7 +15,7 @@ class BazelProcess internal constructor(
     private val originId: String?
 ) {
 
-  fun waitAndGetResult(ensureAllOutputRead: Boolean = true): BazelProcessResult {
+    fun waitAndGetResult(ensureAllOutputRead: Boolean = false): BazelProcessResult {
         val stopwatch = Stopwatch.start()
         val outputProcessor: OutputProcessor =
           if (ensureAllOutputRead) SyncOutputProcessor(process, logger::message, LOGGER::info)
--- a/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.kt
+++ b/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.kt
@@ -3,6 +3,8 @@ package org.jetbrains.bsp.bazel.bazelrunner
 import java.time.Duration
 import org.apache.logging.log4j.LogManager
 import org.jetbrains.bsp.bazel.bazelrunner.outputs.AsyncOutputProcessor
+import org.jetbrains.bsp.bazel.bazelrunner.outputs.OutputProcessor
+import org.jetbrains.bsp.bazel.bazelrunner.outputs.SyncOutputProcessor
 import org.jetbrains.bsp.bazel.commons.Format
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspCompilationManager.java
```diff
@@ -27,7 +27,7 @@ public BepBuildResult buildTargetsWithBep(TargetsSpec targetSpecs, Seq<String> e
             .withFlags(extraFlags.asJava())
             .withTargets(targetSpecs)
             .executeBazelBesCommand()
-            .waitAndGetResult();
+            .waitAndGetResult(false);
```
abrams27, line 30: do we want to wait for the entire output for run & test in the `ExecuteService`?

steveniemitz: sorry I'm not sure what you're asking exactly here?

abrams27: we are calling bazel in `ExecuteService` as well - now with flag set on true, and the question is do we want to use the sync collector there or maybe the async one?

steveniemitz: ah, sorry I misread that as `ExecutorService` and got very confused :D

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.kt
```diff
@@ -13,9 +15,11 @@ class BazelProcess internal constructor(
     private val originId: String?
 ) {
 
-    fun waitAndGetResult(): BazelProcessResult {
-        val outputProcessor = AsyncOutputProcessor(process, logger.withOriginId(originId)::message, LOGGER::info)
+    fun waitAndGetResult(ensureAllOutputRead: Boolean = false): BazelProcessResult {
```
tpasternak, line 18: I would prefer to default to `true` 

cc @abrams27

abrams27 (approved):

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelProcess.kt
abrams27: can be true for me, but then we need to set it to false in the `ExecuteService`

tpasternak: true is just safer

steveniemitz: imo we probably shouldn't be trying to parse the output at all for anything, and use BEP/BES instead.  We could get the execution and workspace roots from the BEP stream as well, and then there'd be no need to have a sync version of this at all.

steveniemitz:
is this good to land?  I'm going to follow up with another review that uses BEP to get the workspace and execution roots, which will let us remove the sync version completely.

abrams27:
yea, sorry i forgot about this PR

PR #289, opened 2022-08-18
[change] pass originId into server responses

azdrojowa123:
Additionally I added `BazelFlag.color(true)` for building target (useful in `intellij-bsp` plugin).

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectSyncService.java
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectSyncService.java
@@ -103,12 +103,12 @@ public class ProjectSyncService {
   }
 
   public ScalaTestClassesResult buildTargetScalaTestClasses(ScalaTestClassesParams params) {
-    var project = projectProvider.get();
+    var project = projectProvider.get(params.getOriginId());
     return bspMapper.buildTargetScalaTestClasses(project, params);
   }
 
   public ScalaMainClassesResult buildTargetScalaMainClasses(ScalaMainClassesParams params) {
-    var project = projectProvider.get();
+    var project = projectProvider.get(params.getOriginId());
     return bspMapper.buildTargetScalaMainClasses(project, params);
   }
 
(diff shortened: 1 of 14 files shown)


server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectSyncService.java
```diff
@@ -103,12 +103,12 @@ public ScalacOptionsResult buildTargetScalacOptions(ScalacOptionsParams params)
   }
 
   public ScalaTestClassesResult buildTargetScalaTestClasses(ScalaTestClassesParams params) {
-    var project = projectProvider.get();
+    var project = projectProvider.get(params.getOriginId());
```
abrams27, line 106: do we need this?

abrams27 (commented):
im not sure about passing the origin id to the project storage, but it seems useful since we are logging there as well - @lukaszwawrzyk what do u think?

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectSyncService.java
azdrojowa123: for our purpose probably not, but due to the fact that `ScalaTestClassesParams` has originId I passed them on

lukaszwawrzyk:
> im not sure about passing the origin id to the project storage, but it seems useful since we are logging there as well - @lukaszwawrzyk what do u think?

We are loging it because automagic error reporting is broken in IntelliJ duriing build, but not during sync. Maybe new bsp plugin could fix that. Maybe we can keep ProjectStorage as it was before. The api is confusing now, Initially I though that project is cached by origin id now or something and it is just error logging. 

> Additionally I added `BazelFlag.color(true)` for building target (useful in `intellij-bsp` plugin).

Sadly it breaks the original scala plugin with bsp support as its build console doesn't support colors :/

To solve this we might check who is our bsp client, and if it is the new plugin we can color the output.


-------------

Another thing. There is a lot of context passing (originId) that makes it harder to see what is important. It should be possible to refactor. Not sure if it can be perfect, but some places would be more clean.
How about adding a method on BspClientLogger like `withOrigin(originId)`. It would return a new logger object which logs everything under this origin id, so we don't have to pass it to each method call anymore. You could just pass such adjusted logger in some places instead of passing in origin id through couple of methods so that it can be used in the logger call. I think this idea might be taken further, but it might be good enough.

lukaszwawrzyk (approved):

abrams27 (approved):

PR #292, opened 2022-08-26
[fix] originId in build is nullable

azdrojowa123:


server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ExecuteService.kt
```diff
@@ -73,7 +73,7 @@ class ExecuteService(
         return CleanCacheResult(bazelResult.stdout, true)
     }
 
-    private fun build(bspIds: List<BuildTargetIdentifier>, originId: String): BazelProcessResult {
+    private fun build(bspIds: List<BuildTargetIdentifier>, originId: String?): BazelProcessResult {
```
azdrojowa123, line 76: in API originId is nullable so in case of not sending originId an error has occurred  :/

abrams27 (approved):

PR #293, opened 2022-08-29
[change] improve reglobbing logic for bloop reglobbing

steveniemitz:
This improves the logic in the reglobber to be smarter if multiple targets share the same base path, for example, if multiple targets are in the same BUILD file in `root`, and have targets that look like:

```
scala_library(
  name = 'a',
  srcs = [globs("a/src/main/**.scala")]
)

scala_library(
  name = 'b',
  srcs = [globs("b/src/main/**.scala")]
)
```

Previously the logic would end up with both `a` and `b` having a glob of `**.scala`, the new logic in this review changes it so a would be a glob with root directory of `root/a/src/main` and b of `root/b/src/main`.

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/ReGlobber.kt
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/ReGlobber.kt
@@ -27,10 +27,10 @@ object ReGlobber {
             return basePath
         }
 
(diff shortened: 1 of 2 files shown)


abrams27 (commented):
one of bloop related tests is failing

steveniemitz:
It actually caught a bug!  I was fixing it up after I submitted this then got distracted.  I'll push the fix later this morning.

abrams27 (approved):

PR #298, opened 2022-09-19
[fix] use source deps from JavaInfo if they are available for thrift targets

steveniemitz:
Certain thrift generators (scrooge) now put the library source directly on the thrift_library target.  This changes the language plugin to also look at the JavaInfo on the target (if its populated) and get the sources from there as well.


abrams27 (changes requested):
thanks a lot! lil comments and please update the changelog as well

abrams27 (approved):
perfect, thanks a lot!

PR #300, opened 2022-09-28
[fix] postpone bazel info

tpasternak:
Before this change, `bazel info` was called before the actual BSP
server. This means, there was no clear way to submit failure (for
example if `bazel` executable was not available in PATH). Currently,
it is run on `build/intitialize` request, so the error will be logged
via BSP failure response.


abrams27 (approved):
cool, thanks a lot!

PR #306, opened 2022-10-25
[fix] adding determining language type for jvm binary targets with no sources

sellophane:


server/src/main/java/org/jetbrains/bsp/bazel/server/sync/BazelProjectMapper.kt
```diff
@@ -89,15 +89,22 @@ class BazelProjectMapper(
         target.dependenciesList.map { Label(it.id) }
 
     private fun inferLanguages(target: TargetInfo): Set<Language> =
-        target.sourcesList.flatMap { source: FileLocation ->
-            Language.all().filter { isLanguageFile(source, it) }
-        }.toHashSet()
+        if (target.sourcesList.isEmpty()) {
+            Language.all().filter { isBinaryTargetOfLanguage(target.kind, it) }.toHashSet()
```
abrams27, line 93: i think that we can actually call it alway, and for libraries and tests as well

sellophane: Libraries and tests both have source files, so sources-based language detecting works better for them (especially for multi-language ones).

CHANGELOG.md
```diff
@@ -28,12 +28,14 @@
 - `JVMLanguagePluginParser.calculateJVMSourceRoot` does not throw 
   an exception for package longer than the path.
   | [#294](https://github.com/JetBrains/bazel-bsp/pull/294)
-- Fix transitive target failure check in bloop export.
+- Fix transitive target failure check in bloop export.aa
```
abrams27, line 31:
```suggestion
- Fix transitive target failure check in bloop export.
```

CHANGELOG.md
```diff
@@ -28,12 +28,14 @@
...
   | [#287](https://github.com/JetBrains/bazel-bsp/pull/287)
 - Use direct source dependencies for thrift targets if present.
   | [#298](https://github.com/JetBrains/bazel-bsp/pull/298)
 - Postpone bazel info.
   | [#300](https://github.com/JetBrains/bazel-bsp/pull/300)
+- Correctly determine language type for JVM binary targets.
```
abrams27, line 37: can u move it to the "unreleased" section?

abrams27 (changes requested):

abrams27 (approved):

PR #309, opened 2022-11-14
[feature] adding test package to JvmTestEnvironment data

sellophane:


sellophane:
> IMO it would be cleaner to add a field in protocol, instead of a special-purpose environment variable

https://github.com/build-server-protocol/build-server-protocol/pull/380

abrams27:
hmmmmm, it's kinda hacky, maybe we can try to move stuff in the protocol

abrams27 (approved):

PR #312, opened 2022-12-21
[fix] handle java packages that does not match directory structure

tpasternak:
Previously, the `calculateJVMSourceRoot` function was falling back to parent of the `source` path only in case the package name was longer than the directory hierarchy. But in case the package mismatched the directory structure, but if it was shorter than directory hierarchy length, it returned incorrect value.


abrams27 (commented):
please update changelog and we can merge

abrams27 (approved):

Issue #92, opened 2021-08-25
"Invalid version format" when using Bazel rolling releases


philwo:
Hi,

I tried to use the Bazel BSP feature in IntellIJ IDEA 2021.2.1 (UE) with Scala plugin 2021.2.17 and Bazel itself as the project. 

When I use Bazel rolling releases, the plug-in crashes with an "Invalid version format" exception in this line: https://github.com/JetBrains/bazel-bsp/blob/dev/bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/data/SemanticVersion.java#L34

I think this is because the regex used rejects the version number. Maybe you could use the regex given in the SemVer 2.0 spec to parse the string instead? See: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

Details:

```
$ export USE_BAZEL_VERSION=rolling
$ bazel version
Bazelisk version: development
Build label: 5.0.0-pre.20210817.2
Build target: bazel-out/darwin_arm64-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Aug 25 11:16:44 2021 (1629890204)
Build timestamp: 1629890204
Build timestamp as int: 1629890204
```

Stack trace:

```
java.lang.IllegalArgumentException: Invalid version format
	at org.jetbrains.bsp.bazel.server.bazel.data.SemanticVersion.<init>(SemanticVersion.java:34)
	at org.jetbrains.bsp.bazel.server.bazel.data.SemanticVersion.fromReleaseData(SemanticVersion.java:56)
	at org.jetbrains.bsp.bazel.server.bazel.data.BazelData.<init>(BazelData.java:21)
	at org.jetbrains.bsp.bazel.server.bazel.BazelDataResolver.resolveBazelData(BazelDataResolver.java:30)
	at org.jetbrains.bsp.bazel.server.BazelBspServer.<init>(BazelBspServer.java:48)
	at org.jetbrains.bsp.bazel.server.ServerInitializer.main(ServerInitializer.java:49)
```

philwo:
Example of the current regex not fully matching the version number of Bazel rolling releases: https://regex101.com/r/ODrInw/1

Example of the official SemVer regex correctly matching it: https://regex101.com/r/sIqC4o/1/

PR #318, opened 2022-12-28
Support Bazel 6.0.0

tpasternak:
1. Main Repository Target Label Prefixes.
In Bazel 5 and older, the target labels were strigified to "//"-prefixed names. Since
6.0.0, they are stringified to "@//"-prefixed names. Unfortunately, this does not apply
to data coming from BEP. Targets references in BEP events payload are "//"-prefixed.
For this reason we need to add "@" manually after retrieving data from BEP.

2. aspects.bzl: Retrieve runtime toolchain in the new way (Bazel 6.0.0)

3. cpp-project: Old GTest version is incompatible with Bazel 6.0.0, the
new one required cpp version bump (--std=c++17)

4. e2e tests are now adapted to run across multiple Bazel versions (currently 5 and 6)

5. JdkResolver: JDK version was inferred from source version, which is not always true. That's
why it is safer just to pass null.

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JdkResolver.kt
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JdkResolver.kt
@@ -59,13 +59,12 @@ class JdkResolver(
           null
     val javaHome = javaHomeFile?.let { bazelPathsResolver.resolveUri(it) }
 
-    val version = null
-    return JdkCandidateData(hasRuntimeJavaHome, javaHome, version)
+    return JdkCandidateData(hasRuntimeJavaHome, javaHome)
         .takeIf { javaHome != null }
   }
 
   private inner class JdkCandidate(private val data: JdkCandidateData) {
-    val version = data.version ?: data.javaHome?.let { jdkVersionResolver.resolve(it.toPath()) }?.toString()
+    val version = data.javaHome?.let { jdkVersionResolver.resolve(it.toPath()) }?.toString()
     val javaHome by data::javaHome
     val isRuntime by data::isRuntime
     val isComplete = javaHome != null && version != null
@@ -75,7 +74,6 @@ class JdkResolver(
   private data class JdkCandidateData(
       val isRuntime: Boolean,
       val javaHome: URI?,
-      val version: String?
   )
 
 }


lukaszwawrzyk (approved):

bazelrunner/src/main/java/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
```diff
@@ -1,16 +1,31 @@
...
+) {
+  val major: Int by lazy {
+    """(?<=release )\d+(?=[0-9.]*)""".toRegex().find(versionString)?.value?.toInt()!!
+  }
+
+  fun mainRepositoryReferencePrefix() = when(major) {
+    in 0..3 -> throw RuntimeException("Unsupported Bazel version, use Bazel 4 or newer")
+    in 4..5 -> "//"
+    else -> "@//"
+  }
```
abrams27, line 23: can we have tests for it? including this issue as well: https://github.com/JetBrains/bazel-bsp/issues/92

e2e/src/main/java/org/jetbrains/bsp/bazel/base/BazelBspTestBaseScenario.kt
```diff
@@ -9,6 +9,9 @@ import java.nio.file.Paths
 import kotlin.system.exitProcess
 
 abstract class BazelBspTestBaseScenario {
+    val targetPrefix = Runtime.getRuntime().exec("bazel version").inputStream.bufferedReader().readText()
+            .let {"""(?<=Build label: )\d+(?=[0-9.]+)""".toRegex().find(it)!!.value.toInt() }
```
abrams27, line 13: do we need triple " here?

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/languages/java/JdkResolver.kt
```diff
@@ -59,14 +59,9 @@ class JdkResolver(
           null
     val javaHome = javaHomeFile?.let { bazelPathsResolver.resolveUri(it) }
 
-    val version =
-        if (targetInfo.hasJavaToolchainInfo())
-          targetInfo.javaToolchainInfo.sourceVersion
-        else
-          null
-
+    val version = null
```
abrams27, line 62: now when the version is null the mechanizm in `JdkCandidate` for the `version` field should be updated (in general we can remove `version` from `JdkCandidateData`)

abrams27 (commented):
i feel that we can do it without bazel version - the bazel version is used only in 3 places:
- `isWorkspaceTarget` - we can just check does the id starts with `//` or `@//`
- similar for extracting the relative path

but then we could do it without passing `BazelInfo` around, what d u think about it?

abrams27 (approved):

PR #322, opened 2023-01-23
[feature] exclude bazel-* symlinks

lukaszwawrzyk:
I implemented the unused endpoint outputPaths.
 
It would now be possible to add outputs for each module, though it doesn't make much sense for bazel, but maybe some day we will find a use for it.

Currently are outputs are treated as directories, it can be changed easily when we find another use case for this endpoint that would need files.

I am detecting if workspace root will be visible in IntelliJ (either whole workspace is imported or bsp server is setup at workspace root), if so, we return a synthetic module at workspace root with single resource root pointing to the workspace and 4 output directories pointing to the problematic symlinks.

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/IntelliJProjectTreeViewFix.kt
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/IntelliJProjectTreeViewFix.kt
@@ -13,6 +13,9 @@ import java.nio.file.Paths
 import kotlin.io.path.Path
 import kotlin.io.path.name
 import kotlin.io.path.toPath
+import kotlin.io.path.listDirectoryEntries
+import kotlin.io.path.isSymbolicLink
+import kotlin.io.path.readSymbolicLink
 
 class IntelliJProjectTreeViewFix(
     private val bazelPathsResolver: BazelPathsResolver,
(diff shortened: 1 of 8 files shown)


lukaszwawrzyk:
I need to fix exec root symlink name inference, it doesn't work with older bazel.

server/src/main/java/org/jetbrains/bsp/bazel/server/sync/IntelliJProjectTreeViewFix.kt
```diff
@@ -9,29 +10,77 @@ import org.jetbrains.bsp.bazel.workspacecontext.WorkspaceContext
...
+            createWorkspaceRootModule(workspaceRoot, modules)
+        } else {
+            createMultipleModules(workspaceContext, workspaceRoot, modules)
+        }
     }
 
-    private fun isFullWorkspaceImport(workspaceContext: WorkspaceContext): Boolean =
-        importTargetSpecs(workspaceContext).any { it.startsWith("//...") }
+    private fun isFullWorkspaceImport(workspaceContext: WorkspaceContext, workspaceRoot: URI): Boolean {
+        return importTargetSpecs(workspaceContext).any { it.startsWith("//...") } ||
```
abrams27, line 38:
```suggestion
        return importTargetSpecs(workspaceContext).any { it.startsWith("//...") || it.startsWith("@//...") } ||
```

we need it for bazel 6 i think

abrams27 (approved):

PR #347, opened 2023-02-24
[change] Don't request the "bsp-ide-resolve-transitive-deps" output group when running the bsp_target_info_aspect.

odisseus:
Upon receiving the `workspace/buildTargets` request, the server runs a Bazel aspect to retrieve the data about the project structure. We used to request two output groups, namely `bsp-target-info-transitive-deps` and `bsp-ide-resolve-transitive-deps`. It turns out that the files produced by the latter one are never used, so we might as well drop it.

This change should greatly speed up importing fresh projects, because requesting `bsp-ide-resolve-transitive-deps` causes source jars to be built. On a sample project with 10 000 empty Java targets, the duration of fresh import has decreased from about 300 seconds to less than 90 seconds.

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
@@ -50,13 +50,12 @@ class ProjectResolver(
         bazelBspAspectsManager.fetchFilesFromOutputGroups(
             workspaceContext.targets,
             ASPECT_NAME,
-            listOf(BSP_INFO_OUTPUT_GROUP, ARTIFACTS_OUTPUT_GROUP)
+            listOf(BSP_INFO_OUTPUT_GROUP)
         )
 
 
     companion object {
         private const val ASPECT_NAME = "bsp_target_info_aspect"
         private const val BSP_INFO_OUTPUT_GROUP = "bsp-target-info-transitive-deps"
-        private const val ARTIFACTS_OUTPUT_GROUP = "bsp-ide-resolve-transitive-deps"
     }
 }


abrams27:
does it mean that we can remove the creation of the group in the aspect file? https://github.com/JetBrains/bazel-bsp/search?q=bsp-ide-resolve
also, please update the changelog

lukaszwawrzyk:
We can't just remove remove it. Scope of the build can be reduced and the build of it can be postponed.
This contains not only source jars, but actual jars for 3rd party libraries and dependencies from within the project.

abrams27 (approved):

odisseus:
I think we actually can remove large parts of the Bazel aspect. The method `buildProjectWithAspect` is [called](https://github.com/JetBrains/bazel-bsp/blob/f5bd4c42782a7e328f19fc54ef4f6becce22dc0e/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt#L30-L33) only in one place, and it doesn't look like it makes any use of the other output groups.

However, I still don't understand completely how the aspect works, and whether the server code  could implicitly rely on the existence of its other outputs.

lukaszwawrzyk:
> I think we actually can remove large parts of the Bazel aspect. The method `buildProjectWithAspect` is [called](https://github.com/JetBrains/bazel-bsp/blob/f5bd4c42782a7e328f19fc54ef4f6becce22dc0e/server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt#L30-L33) only in one place, and it doesn't look like it makes any use of the other output groups.
> 
> However, I still don't understand completely how the aspect works, and whether the server code could implicitly rely on the existence of its other outputs.

The aspect for example finds outputs of compilation of each imported target. Their paths are stored in bsp-info proto files. Proto files are created by bsp-target-info-transitive-deps group. The actual jars are built by bsp-ide-resolve-transitive-deps group. If we do not request anything from the second group at all, we will only know where the jars *would* be if they were built.

abrams27:
closing the ticket - we are planning to introduce the same change in the 3.0. - we will do a sync -> build -> resync flow to collect all the jars we need

PR #348, opened 2023-02-24
Add an empty WORKSPACE file to the directory in which the distributable aspects are stored.

odisseus:
This is useful for testing the changes in another project (using `--override_repository`) without having to re-install the entire BSP server in that project's base directory.


abrams27:
if it's useful for u we can merge it, but it is possible to edit the aspect in the destination project to test the changes

PR #372, opened 2023-04-05
[fix] Add class jars generated during annotation processing

tpasternak:
It would also be possible to pass source jars, but there's no clear way how it should be handled in BSP

https://bazel.build/rules/lib/JavaInfo#JavaInfo


abrams27 (commented):
dont forget to update the changelog

abrams27 (approved):

PR #377, opened 2023-04-07
[fix] set PublishDiagnosticsParams.reset to be true | #BAZEL-375 Done

tanishiking:
fix https://youtrack.jetbrains.com/issue/BAZEL-375


abrams27 (approved):
thanks a lot!

PR #402, opened 2023-05-09
[fix] Fixup failed target names from BEP in bazel 6+

steveniemitz:
Target names in BEP in bazel 6 aren't prefixed with `@`, which causes a mismatch in comparing the failed targets with project targets.

### DIFF
--- a/server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporter.java
+++ b/server/src/main/java/org/jetbrains/bsp/bazel/server/bloop/BloopExporter.java
@@ -19,6 +19,7 @@ import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.eclipse.lsp4j.jsonrpc.CancelChecker;
+import org.jetbrains.bsp.bazel.bazelrunner.BazelInfo;
 import org.jetbrains.bsp.bazel.bazelrunner.BazelRunner;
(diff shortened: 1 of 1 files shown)


abrams27 (approved):
thanks a lot, please update the changelog

i have also another thing to u: we were thinking that this is not the best place for bloop support - we dont know bloop well (so it's difficult to maintain it), and it actually doesnt use that much from our codebase - the bazel-bsp server doesnt even exist in the bloop mode;

since bazel-bsp can be used as a maven dependency, we have 2 propositions:
- move it to a completely separate repo (not under jetbrains)
- move it to the bloop itself
what do u think about it?

steveniemitz:
> move it to a completely separate repo (not under jetbrains)

I could get behind this.  Agreed its not really sharing much from bazel-bsp that couldn't be just exposed as APIs.

abrams27:
perfect, then we will deprecate it in the following release, we are planning to release 3.0 with breaking changes around bsp notifications and project view in next ~2 months 

can i merge this pr?

steveniemitz:
yeah this is good to merge.  It looks like the google-format failure was unrelated to this change.

Deprecation for 3.0 seems fine to me, I'll figure out a better place for it.

abrams27:
fyi: we are planning to do the 3.0 release next week (or the following one) - it will also remove bloop from bazel-bsp

cc @lukaszwawrzyk

PR #411, opened 2023-05-15
[maintenance] add grouping for dependencies that share version in Bazel Steward

lukaszwawrzyk:
This will use new Bazel Steward feature that creates one PR for group of dependencies.


abrams27 (approved):
wooo, nice!! thanks

PR #420, opened 2023-06-01
[fix] prioritize most frequently used JDKs when selecting the project JDK

lukaszwawrzyk:
Currently JDKs are sorted lexicographically by java home for determinism. Sometimes it can be a problem. when there are 2 JDKs found and one of them is used in nearly all targets and one is used in a few docker related targets that are not relevant. If alphabet is not on your side, you get the wrong JDK (for example wrong architecture) that is not even fetched/resolved fully. 
Since JDK resolving is still a heuristic in bazel-bsp, this change should give more useful results usually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 
 ## [Unreleased]
+### Fixes 🛠️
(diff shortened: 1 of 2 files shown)


abrams27 (approved):
thx

PR #425, opened 2023-06-06
[fix] set uncaught execution handlers in executor services

tpasternak:
Otherwise, an executor service could crash (for example due to OutOfMemory exception) without calling process exit, so as a result the process would hang.


abrams27 (approved):
thanks! dont forget about the changelog

PR #429, opened 2023-06-13
[performance] Large synthetic project benchmark 

tpasternak:
Prepare benchmark


server/src/main/java/org/jetbrains/bsp/bazel/server/bep/BepOutputBuilder.java
```diff
@@ -32,6 +34,12 @@ public void storeTargetOutputGroups(String target, List<OutputGroup> outputGroup
   }
 
   public BepOutput build() {
+    System.gc();
+    LOGGER.info(
```
abrams27, line 38: it feels more like debug i think

abrams27 (approved):

PR #434, opened 2023-06-16
[maintenance] Dump log metrics to file

tpasternak:
Example run:
`bazel run //bspcli -- <path-to-project> $PWD/metrics.txt`


server/src/main/java/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
```diff
@@ -2,34 +2,47 @@ package org.jetbrains.bsp.bazel.server.sync
...
+        private val workspaceContextProvider: WorkspaceContextProvider,
+        private val bazelProjectMapper: BazelProjectMapper,
+        private val bspLogger: BspClientLogger,
+        private val targetInfoReader: TargetInfoReader,
+        private val bazelInfo: BazelInfo,
+        private val metricsLogger: MetricsLogger?
 ) {
+    private fun <T> measured(description: String, f: () -> T): T {
+        val sw = Stopwatch.start()
+        val result: T =  f()
```
abrams27, line 27:
```suggestion
        val result = f()
```

abrams27 (approved):

PR #435, opened 2023-06-21
[maintenance] disable updating bloop in Bazel Steward

lukaszwawrzyk:


abrams27:
no need for that, we will remove all the depdencies related to bloop as well

PR #454, opened 2023-08-24
[maintenance] e2e tests now use `bazel_integration_rules`

abrams27:
now our e2e will be executed using [dedicated rules](https://github.com/bazel-contrib/rules_bazel_integration_test)  - it gives us more flexibility, and we dont have to maintain the custom test execution

in the following prs i'll:
- fix the issue with system specific things
- try to use junit in the tests / split checks

TC e2e test will fail since the config is outdated (i changed it manually for one single repo test only)


hb-man (approved):
Okay, TC configs looks fine. Adding it to internal TC config on buildserver before merge.

PR #455, opened 2023-08-25
[fix] now server can obtain scala3 compiler

abrams27:
https://youtrack.jetbrains.com/issue/BAZEL-404/Bazel-Bsp-with-scala-3.1.0-Failed-to-resolve-Scala-SDK-for-project

the solution is kinda ugly, but we will be changing the lang support soon and then we will have to handle all the scala 3 things more gracefully

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@
 
 ### Fixes 🛠️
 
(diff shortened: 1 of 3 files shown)


xuansontrinh (approved):
LGTM 👍

PR #512, opened 2024-01-17
[feature] Add an option to override detected rules 

tgodzik:
The check here was a bit too fragile. If rules_scala were add in another function then generator_function would have that function as a value, which means we would not discover the Scala aspect.

I managed to get everything working in a quite a large repo after the change. We also have later a check so this should not cause any issues.


xuansontrinh:
Thank you for your contribution.
As far as I remember, the reason why I enforced such a tight check is to avoid false positives when there are transitive rules, which are not available for aspects to use during the analysis phase.
I will need to check this against some repos to see if it is fine to loosen the check.

xuansontrinh:
Thank you for your contribution.
As far as I remember, the reason why I enforced such a tight check is to avoid false positives when there are transitive rules, which are not available for our aspects to use during the analysis phase.
I will need to check this against some repos to see if it is fine to loosen the check.

tgodzik:
> Thank you for your contribution. As far as I remember, the reason why I enforced such a tight check is to avoid false positives when there are transitive rules, which are not available for our aspects to use during the analysis phase. I will need to check this against some repos to see if it is fine to loosen the check.

Were it still needed, maybe we could have an alternative check since this doesn't allow for much flexibility in the build scripts

xuansontrinh:
what is the value of of `value` field in the corresponding rule in your example?
The check that I had already considered the case where you wrap rules inside a generator, which calls `http_archive`.
If possible, could you please provide the minimal example that reproduces the problem? I can think of a way to loosen the check but also make sure it's not too loose.

Regarding the alternative solution, as far as I am aware, there is no more effective check than the current one with the old WORKSPACE-based external dependency system. In case you know of some other way, please let me know. I am happy to support.

lukaszwawrzyk:
As a quick reproduction, create `languages` macro in .bzl file, import it and call it from WORKSPACE. The macro should use `maybe` to create an `http_archive` with rules_scala.

On the sidenote, I think it is very important to have languages configurable to override this mechanism. For one, people could save time on bazel call that detects them and more importantly, only use languages that they want. For example, having rust in bazel repo doesn't mean we want to work with rust. It can even be outside the import scope.

xuansontrinh:
@lukaszwawrzyk I think adding it as a custom setting can be a way.
I will discuss this with the team and see what we can do about it.

tgodzik:
Any updates? Would be great to make it work, if a setting is better I can modify the PR, though would be great to get some pointers in that case

xuansontrinh:
@tgodzik So we agree that it is necessary to have a configurable field in projectview file to override the current rule detection mechanism.
For that, I created [a ticket](https://youtrack.jetbrains.com/issue/BAZEL-817/Configurable-enabled-rules-list-in-projectview-file) with some implementation details to guide you through.
Please let me know if you have any questions about the implementation.

tgodzik:
Thanks! I will take a look.

tgodzik:
@xuansontrinh is this the right way? I will need to work a bit more on the tests still and I will try to reproduce the issue in e2e tests if the current approach is ok.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelExternalRulesQuery.kt
```diff
@@ -41,8 +49,7 @@ class BazelWorkspaceExternalRulesQueryImpl(private val bazelRunner: BazelRunner)
   private fun Document.calculateEligibleRules(): List<String> {
     val xPath = XPathFactory.newInstance().newXPath()
     val expression =
-      "/query/rule[contains(@class, 'http_archive') and " +
-        "(not(string[@name='generator_function']) or string[@name='generator_function' and contains(@value, 'http_archive')])" +
```
xuansontrinh, line 45: I would prefer to revert this change as it will introduce a lot of false positives.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/common/ServerContainer.kt
```diff
@@ -79,9 +75,11 @@ class ServerContainer internal constructor(
       val bazelProjectMapper =
         BazelProjectMapper(languagePluginsService, bazelPathsResolver, targetKindResolver, bazelInfo, bspClientLogger, metricsLogger)
       val targetInfoReader = TargetInfoReader()
+      val currentContext = workspaceContextProvider.currentWorkspaceContext()
+      val bazelExternalRulesQuery = if (currentContext.enabledRules.values.isEmpty()) BazelExternalRulesQueryImpl(bazelRunner, bazelInfo.isBzlModEnabled) else BazelEnabledRulesQueryImpl(currentContext.enabledRules)
```
xuansontrinh, line 79: please put the implementation part inside `BazelExternalRulesQueryImpl#fetchExternalRules`.
You can pass additional argument `enabledRules` to `BazelExternalRulesQueryImpl` also.

executioncontext/projectview/src/main/kotlin/org/jetbrains/bsp/bazel/projectview/model/ProjectView.kt
```diff
@@ -1,37 +1,30 @@
...
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewBazelBinarySection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewBuildFlagsSection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewBuildManualTargetsSection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewDeriveTargetsFromDirectoriesSection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewDirectoriesSection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewExcludableListSection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewImportDepthSection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewListSection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewSingletonSection
-import org.jetbrains.bsp.bazel.projectview.model.sections.ProjectViewTargetsSection
```
xuansontrinh, line 13: Please revert wildcard imports.
We prefer to use single line imports in the codebase.

xuansontrinh (changes requested):

executioncontext/workspacecontext/src/main/kotlin/org/jetbrains/bsp/bazel/workspacecontext/WorkspaceContext.kt
```diff
@@ -63,6 +63,12 @@ data class WorkspaceContext(
      * Obtained from `ProjectView` simply by mapping `import_depth` section.
      */
     val importDepth: ImportDepthSpec,
+        /**
```
xuansontrinh, line 66: format error here.

executioncontext/projectview/src/main/kotlin/org/jetbrains/bsp/bazel/projectview/model/ProjectView.kt
```diff
@@ -1,37 +1,30 @@
...
-    /** flag for building manual targets. */
+        /** flag for building manual targets. */
     val buildManualTargets: ProjectViewBuildManualTargetsSection?,
-    /** directories included and excluded from the project  */
+        /** directories included and excluded from the project  */
     val directories: ProjectViewDirectoriesSection?,
-    /** if set to true, relevant project targets will be automatically derived from the `directories` */
+        /** if set to true, relevant project targets will be automatically derived from the `directories` */
     val deriveTargetsFromDirectories: ProjectViewDeriveTargetsFromDirectoriesSection?,
-    /** level of depth for importing inherited targets */
```
xuansontrinh, line 33: is there a reason for this additional indentation?
This looks weird to me.

xuansontrinh (changes requested):
this PR requires some changes.
But beside that, it is heading in the right direction.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelExternalRulesQuery.kt
xuansontrinh, line 45: Please revert this change as it will introduce a lot of false positives.

xuansontrinh (changes requested):

e2e/src/main/kotlin/org/jetbrains/bsp/bazel/BazelBspScalaProjectTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import ch.epfl.scala.bsp4j.ScalacOptionsItem
+import ch.epfl.scala.bsp4j.ScalacOptionsParams
+import ch.epfl.scala.bsp4j.ScalacOptionsResult
+import org.jetbrains.bsp.bazel.base.BazelBspTestBaseScenario
+import org.jetbrains.bsp.bazel.base.BazelBspTestScenarioStep
+
+import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
+
+object BazelBspScalaProjectTest : BazelBspTestBaseScenario() {
```
tgodzik, line 19: Trying to make this work currently, but getting:
```
ERROR: error loading package '': Every .bzl file must have a corresponding package, but '//utils:tools.bzl' does not have one. Please create a BUILD file in the same or any parent directory. Note that this BUILD file does not need to do anything except exist.
```
and it has the BUILD file

executioncontext/projectview/src/main/kotlin/org/jetbrains/bsp/bazel/projectview/model/ProjectView.kt
tgodzik: Sure, it got done automatically, wanted to revert that before merging

e2e/src/main/kotlin/org/jetbrains/bsp/bazel/BazelBspScalaProjectTest.kt
xuansontrinh: did you try locking the version for Bazel and test locally (by running this as a standalone Bazel project)?

tgodzik: Looks like files are not copied correctly, at some point it stopped finding WORKSPACE file as well :/

xuansontrinh: @abrams27 maybe you know about this issue?

tgodzik: It seems that only the main directory is copied, any subdirectories are not :/ I don't really understand how is that possible. Works correctly for the python one on the same machine

tgodzik: Figured it out, it needs to be added to https://github.com/JetBrains/bazel-bsp/blob/master/.bazelrc#L9 <- any chance to fix that requirement?

xuansontrinh: Noted.
I think we can later write detailed instructions for adding a new e2e test for this repo.
Or auto-generate `.bazelrc` to cater for the tests.
We will see about that.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/common/ServerContainer.kt
tgodzik: Done!

executioncontext/workspacecontext/src/main/kotlin/org/jetbrains/bsp/bazel/workspacecontext/WorkspaceContext.kt
tgodzik: Fixed

executioncontext/projectview/src/main/kotlin/org/jetbrains/bsp/bazel/projectview/model/ProjectView.kt
tgodzik: That was automatic, removed.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelExternalRulesQuery.kt
tgodzik: Reverted!

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelExternalRulesQuery.kt
```diff
@@ -13,18 +14,32 @@ interface BazelExternalRulesQuery {
...
 class BazelExternalRulesQueryImpl(
   private val bazelRunner: BazelRunner,
-  private val isBzlModEnabled: Boolean
+  private val isBzlModEnabled: Boolean,
+  private val enabledRules: EnabledRulesSpec
 ) : BazelExternalRulesQuery {
-  override fun fetchExternalRuleNames(cancelChecker: CancelChecker): List<String> = when (isBzlModEnabled) {
-    true -> BazelBzlModExternalRulesQueryImpl(bazelRunner).fetchExternalRuleNames(cancelChecker) +
-      BazelWorkspaceExternalRulesQueryImpl(bazelRunner).fetchExternalRuleNames(cancelChecker)
+    override fun fetchExternalRuleNames(cancelChecker: CancelChecker): List<String> {
```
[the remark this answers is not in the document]
tgodzik: Changed!

executioncontext/projectview/src/main/kotlin/org/jetbrains/bsp/bazel/projectview/model/ProjectView.kt
```diff
@@ -32,6 +33,8 @@ data class ProjectView(
     val deriveTargetsFromDirectories: ProjectViewDeriveTargetsFromDirectoriesSection?,
     /** level of depth for importing inherited targets */
     val importDepth: ProjectViewImportDepthSection?,
+    /** manually enabled rules to overcome false negatives*/
```
xuansontrinh, line 36: I would change the description to `manually enabled rules to override the automatic rules detection mechanism`.

executioncontext/projectview/src/main/kotlin/org/jetbrains/bsp/bazel/projectview/model/ProjectView.kt
```diff
@@ -184,6 +192,16 @@ data class ProjectView(
         private fun combineImportDepthSection(importedProjectViews: List<ProjectView>): ProjectViewImportDepthSection? =
             importDepth ?: getLastImportedSingletonValue(importedProjectViews, ProjectView::importDepth)
 
+        private fun combineManualRulesSection(importedProjectViews: List<ProjectView>): ProjectViewEnabledRulesSection? {
```
xuansontrinh, line 195: rename to `combineEnabledRulesSection`.

executioncontext/workspacecontext/src/main/kotlin/org/jetbrains/bsp/bazel/workspacecontext/EnabledRulesSpec.kt
```diff
@@ -0,0 +1,24 @@
...
+}
+
+private val defaultEnabledRulesSpec = EnabledRulesSpec(values = emptyList())
+
+internal object EnabledRulesSpecExtractor : ExecutionContextEntityExtractor<EnabledRulesSpec> {
+
+    override fun fromProjectView(projectView: ProjectView): EnabledRulesSpec = when (projectView.enabledRules) {
+        null -> defaultEnabledRulesSpec
+        else -> mapNotEmptySection(projectView.enabledRules!!)
+    }
```
xuansontrinh, line 21: I think it is compact enough to just do this:
```kotlin
EnabledRulesSpec(values = projectView.enabledRules?.values ?: emptyList())
```

xuansontrinh (changes requested):
The PR is in a good state now.
Just some minor cosmetics and documentation change and we can ship it.

install/src/main/kotlin/org/jetbrains/bsp/bazel/install/InstallationContextProvider.kt
```diff
@@ -28,7 +28,9 @@ object InstallationContextProvider {
     fun generateAndSaveProjectViewFileIfNeeded(cliOptions: CliOptions) {
         val generatedProjectViewFilePath = calculateGeneratedProjectViewPath(cliOptions)
         if (!generatedProjectViewFilePath.isFileExisted() || cliOptions.projectViewCliOptions != null) {
-            ProjectViewCLiOptionsProvider.generateProjectViewAndSave(cliOptions, generatedProjectViewFilePath)
+            if (!generatedProjectViewFilePath.exists()) {
```
xuansontrinh, line 31: I think I missed this one: Why do we need to extra guard this?

tgodzik: That was mostly for the tests, but also if `generatedProjectViewFilePath.isFileExisted() ` is true and `cliOptions.projectViewCliOptions != null` is true, this will throw an exception. That popped up in the added e2e test, but also it seems sensible not to try and create a file if it already exists.

xuansontrinh (approved):
I think it's good to go.
Thank you for your contribution.

PR #513, opened 2024-01-18
refactor: Remove CARGO_BAZEL_REPIN flag to make builds faster

tgodzik:
I am still not sure when this should be set, but if we set it here, Bazel will spend additional 30s on any build even if the modules being build do not contain any Rust at all.

Would it make sense to check if the build targets actually contain Rust, is it possible to do?

@tudny I see that it was added by you, any idea how else we could improve a situation.

The context is that we work with a large Bazel codebase, which contains a lot of Scala and Typescript and a much smaller amount of Rust, so having additional 30 s spent on any Typescript or Scala recompilation seems a bit of a productivity killer


abrams27:
this flag is used by the rust part of the code, but we can add it only if the rust rules are detected

lukaszwawrzyk:
This is the whole problem here. Rust rules are detected and dependencies are repinned, which takes extra 20 second on a build of a scala target.
We'd need a way to disable rust. But this is not the only problem.
It is sync's job to update dependencies, not arbitrary build. It adds extra cost to every simple build. It is less than ideal in standard scenarios of building/testing, but it is a real blocker for other editors that use diagnostics to display errors as user types. It adds a big delay.

tgodzik:
Could we add this to sync then? It seems a proper place to resolve the dependencies, otherwise this might be quite hard on the each change.

Or are you 100% convinced this should stay as is? We could try with rules detection then,

abrams27:
yea we can move it to the call in `BazelBspAspectsManager`, in https://github.com/JetBrains/bazel-bsp/pull/512 (or another PR) we can try to add a section to project view file as a "manual" way of enabling supported rules in a project

tgodzik:
> yea we can move it to the call in `BazelBspAspectsManager`, in #512 (or another PR) we can try to add a section to project view file as a "manual" way of enabling supported rules in a project

Moved the call, I will take a look at projectview in the other PR

abrams27 (approved):

PR #517, opened 2024-01-25
[feature] add data to taskStart and taskFinished

jkciesluk:
Metals uses this information for managing compilations. Also moves clearing diagnostics from `targetComplete` to `taskFinished`

### DIFF
--- a/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.java
+++ b/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.java
@@ -1,10 +1,15 @@
 package org.jetbrains.bsp.bazel.server.bep;
 
 import ch.epfl.scala.bsp4j.BuildClient;
+import ch.epfl.scala.bsp4j.BuildTargetIdentifier;
+import ch.epfl.scala.bsp4j.CompileReport;
+import ch.epfl.scala.bsp4j.CompileTask;
 import ch.epfl.scala.bsp4j.StatusCode;
 import ch.epfl.scala.bsp4j.TaskFinishParams;
+import ch.epfl.scala.bsp4j.TaskFinishDataKind;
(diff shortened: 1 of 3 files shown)


server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.java
```diff
@@ -156,9 +170,35 @@ private void consumeFinishedEvent(BuildEventStreamProtos.BuildFinished buildFini
...
+
+  // clear former diagnostics by publishing an empty array of diagnostics
+  // why we do this on `task_finished` instead of `action_completed` or `target_complete`?
+  // - `action_completed` won't be published on build success for a target,
+  // - `target_complete` can be result of build aspect, not compilation.
+  // https://github.com/bazelbuild/bazel/blob/d43737f95d28789bb2d9ef2d7f62320e9a840ab0/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto#L157-L160
+  private void clearDiagnostics() {
+    if (target.isPresent()) {
+      var targetId = target.get().getUri();
+      var label = targetId.startsWith("@") ? targetId.substring(1) : targetId;
```
abrams27, line 196: for bzl mod project it's possible to have `@@` as well, might it be a problem?

abrams27:
just a tiny question, in general look good

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.java
jkciesluk: Turns out we can't clear diagnostics in `taskFinished`, there still were some false positives. I've added it back to `completedEvent`, but with an additional check to ensure we only clear diagnostics on successful compilation

abrams27 (approved):
thanks a lot for the pr!

PR #518, opened 2024-01-29
bugfix: Allow enabled-rules on cli installer

tgodzik:
In the previous PR I forgot to add a proper parsing option. Now, that's fixed and a test is added.


xuansontrinh (approved):
Thank you

PR #520, opened 2024-02-01
bugfix: Cache classpath information in proto files

tgodzik:
Previously, when scalacOptions or javacOptions was queried we would also query each target's classpath, which would take around 0.5 for a target, which means users would need to wait much longer for a fully functional workspace.

Now, we cache the classpath information together with all other jvm information, which makes it all much faster.

This also fixes the issue in https://github.com/JetBrains/bazel-bsp/pull/519

### DIFF
--- a/aspects/rules/jvm/jvm_info.bzl
+++ b/aspects/rules/jvm/jvm_info.bzl
@@ -156,6 +156,8 @@ def extract_jvm_info(target, ctx, output_groups, **kwargs):
         main_class = main_class,
         args = args,
         jdeps = [file_location(j) for j in jdeps],
+        runtime_jars = [jar.path for jar in extract_runtime_jars(target, provider).to_list()],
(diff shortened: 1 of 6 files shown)


tgodzik:
If this is a big issue we could put it under a flag.

abrams27:
so we did it on purpose - classpath computation takes quite some time and memory (for biiigg project quite a lot), and we dont need it during the sync, so we made it a bit more lazy

tgodzik:
> so we did it on purpose - classpath computation takes quite some time and memory (for biiigg project quite a lot), and we dont need it during the sync, so we made it a bit more lazy

The biggest issue I have is that we need both scalacOptions and javacOptions right away. We need to know the Scala version at a minimum for indexing. So in reality we don't need classpath right away also, but we have no other way of getting Scala version.

PR #533, opened 2024-02-29
bugfix: Fix the @@ prefix issue if bazel.module and WORKSPACE are used together

tgodzik:
It's not uncommon to have both BAZEL.module file and WORKSPACE, in that case Bazel BSP will not detect all targets, or in case of the repository I am working on it would detect none of them.

There doesn't seem to be any actual way to discover if a target should have a @ or @@ prefix, so instead I changed the behaviour to replace @@ with @, which should work in most cases.

### DIFF
--- a/bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
+++ b/bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
@@ -16,12 +16,10 @@ data class BazelRelease(
   val major: Int
 ) {
 
-  fun mainRepositoryReferencePrefix(isBzlModEnabled: Boolean) = when (major) {
+  fun mainRepositoryReferencePrefix() = when (major) {
     in 0..3 -> throw RuntimeException("Unsupported Bazel version, use Bazel 4 or newer")
     in 4..5 -> "//"
-    else ->
-      if (isBzlModEnabled) "@@//"
-      else "@//"
+    else -> "@//"
   }
 
   companion object {
--- a/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
+++ b/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
@@ -68,7 +68,7 @@ class ProjectResolver(
     val rootTargets = buildAspectResult.bepOutput.rootTargets().let { formatTargetsIfNeeded(it) }
     val targets = measured(
       "Parsing aspect outputs"
-    ) { targetInfoReader.readTargetMapFromAspectOutputs(aspectOutputs) }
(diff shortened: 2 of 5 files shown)


tgodzik:
Ok, it seems this hasn't broken anything. Opinions? This is also related to https://youtrack.jetbrains.com/issue/BAZEL-863/Bazel-7.x-will-always-try-to-create-module.bazel

agluszak:
LGTM. I don't remember why we decided to use `@@` here in the first place, but for context: https://bazel.build/external/overview#canonical-repo-name

bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
```diff
@@ -16,12 +16,10 @@ data class BazelRelease(
   val major: Int
 ) {
 
-  fun mainRepositoryReferencePrefix(isBzlModEnabled: Boolean) = when (major) {
+  fun mainRepositoryReferencePrefix() = when (major) {
     in 0..3 -> throw RuntimeException("Unsupported Bazel version, use Bazel 4 or newer")
     in 4..5 -> "//"
-    else ->
```
tgodzik, line 22: This was basically a guess previously, since the targets can come with @ and @@ within a single workspace.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
```diff
@@ -68,7 +68,7 @@ class ProjectResolver(
     val rootTargets = buildAspectResult.bepOutput.rootTargets().let { formatTargetsIfNeeded(it) }
     val targets = measured(
       "Parsing aspect outputs"
-    ) { targetInfoReader.readTargetMapFromAspectOutputs(aspectOutputs) }
+    ) { targetInfoReader.readTargetMapFromAspectOutputs(aspectOutputs).let { it.mapKeys { it.key.replace("@@", "@") } } }
```
tgodzik, line 71: Here we instead normalize anything to have single @, which should be safe since this seems to be only a target lookup structure.

abrams27:
with this change it's not possible to import this repo

tgodzik:
> with this change it's not possible to import this repo

How does it break though? I didn't think it would break anything that worked previously :thinking: 

Any other ideas how to fix this? The main problem is that we cannot know if a project should have @ or @@ inside bazel-bsp. Why isn't that saved when exporting via aspect?

tgodzik:
If anyone needs a reproduction then https://github.com/susliko/metals-bazel should behave similarly to the private repo I am working on. We don't get any targets there be it Metals or Intellij.

abrams27:
> How does it break though? I didn't think it would break anything that worked previously

there are places where we reference the id directly, so it's still with `@@` there, but im not 100% sure will it still be enough (but prob it's worth a try)

tgodzik:
Created an alternative approach here https://github.com/JetBrains/bazel-bsp/pull/534 since something weird happened with this PR when I pushed.

PR #534, opened 2024-03-01
[fix] fix the `@@` prefix issue if `bazel.module` and `WORKSPACE` are used together

tgodzik:
Second attempt, this time I changed it to looks for the actual @@ or @ label to be available.

### DIFF
--- a/bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
+++ b/bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
@@ -16,12 +16,16 @@ data class BazelRelease(
   val major: Int
 ) {
 
-  fun mainRepositoryReferencePrefix(isBzlModEnabled: Boolean) = when (major) {
+  fun isRelativeWorkspacePath(label: String) = when (major) {
     in 0..3 -> throw RuntimeException("Unsupported Bazel version, use Bazel 4 or newer")
-    in 4..5 -> "//"
-    else ->
-      if (isBzlModEnabled) "@@//"
-      else "@//"
+    in 4..5 -> label.startsWith("//")
+    else -> label.startsWith("@//") || label.startsWith("@@//")
+  }
+
+  fun stripPrefix(label: String) = when (major) {
+    in 0..3 -> throw RuntimeException("Unsupported Bazel version, use Bazel 4 or newer")
+    in 4..5 -> label.removePrefix("//")
+    else -> label.dropWhile { it == '@' }.removePrefix("//")
   }
 
   companion object {
--- a/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
+++ b/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
@@ -2,6 +2,7 @@ package org.jetbrains.bsp.bazel.server.sync
 
 import org.eclipse.lsp4j.jsonrpc.CancelChecker
 import org.jetbrains.bsp.bazel.bazelrunner.BazelInfo
+import org.jetbrains.bsp.bazel.info.BspTargetInfo
 import org.jetbrains.bsp.bazel.logger.BspClientLogger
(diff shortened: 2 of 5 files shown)


tgodzik:
Ok, so this should be much safer. I don't do any assumptions here about @ or @@. Instead I decided to allow both as both are possible within a single workspace.

bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
```diff
@@ -16,12 +16,16 @@ data class BazelRelease(
...
-    else ->
-      if (isBzlModEnabled) "@@//"
-      else "@//"
+    in 4..5 -> label.startsWith("//")
+    else -> label.startsWith("@//") || label.startsWith("@@//")
+  }
+
+  fun stripPrefix(label: String) = when (major) {
+    in 0..3 -> throw RuntimeException("Unsupported Bazel version, use Bazel 4 or newer")
+    in 4..5 -> label.substring("//".length)
```
abrams27, line 27: i think it's better to use `removePrefix` here

bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
```diff
@@ -16,12 +16,16 @@ data class BazelRelease(
...
+    else -> label.dropWhile { it == '@' }.substring("//".length)
```
abrams27, line 28: and here ofc as well

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
```diff
@@ -55,20 +56,20 @@ class ProjectResolver(
...
+    ) { buildAspectResult.bepOutput.filesByOutputGroupNameTransitive(BSP_INFO_OUTPUT_GROUP) }
+    val targets = measured(
+        "Parsing aspect outputs"
+    ) { targetInfoReader.readTargetMapFromAspectOutputs(aspectOutputs).let { it } }
     val allTargetNames =
       if (buildAspectResult.isFailure)
         measured(
           "Fetching all possible target names"
-        ) { formatTargetsIfNeeded(bazelBspFallbackAspectsManager.getAllPossibleTargets(cancelChecker)) }
+        ) { formatTargetsIfNeeded(bazelBspFallbackAspectsManager.getAllPossibleTargets(cancelChecker), targets) }
```
abrams27, line 69: note: targets from`bazelBspFallbackAspectsManager.getAllPossibleTargets` are not included in `targets so they will be always mapped to `@<target>`, but i think it's not a problem

abrams27 (commented):
just minor comments, but seems to work on our project

bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
tgodzik: Fixed! I couldn't find the method before.

bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/BazelInfo.kt
tgodzik: Done

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt
tgodzik: That should work correctly as far as I can see :thinking:

tgodzik:
I confirmed it just in case on the private repository I am working on and it seems to work alright. I also rebased the PR, which seems to have made the CI issue go away.

tgodzik:
Thanks for taking a look at it!

tgodzik:
We could close https://youtrack.jetbrains.com/issue/BAZEL-863/Bazel-7.x-will-always-try-to-create-module.bazel

PR #536, opened 2024-03-07
feature: Add dependencyModules request support

tgodzik:
This would be super useful to use as an alternative to libraries request, which is not standard BSP request.

Tests depends on https://github.com/build-server-protocol/bsp-testkit2/pull/41

### DIFF
--- a/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapper.kt
+++ b/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapper.kt
@@ -88,6 +88,7 @@ import java.nio.file.Path
 import java.nio.file.Paths
 import java.util.LinkedList
 import kotlin.io.path.exists
+import java.util.*
 import kotlin.io.path.name
 import kotlin.io.path.relativeToOrNull
 import kotlin.io.path.toPath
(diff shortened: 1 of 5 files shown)


tgodzik:
Anyone can take a look at https://github.com/build-server-protocol/bsp-testkit2/pull/41 ? I would be able to add proper tests here. Will try adding unit tests for the time being.

tgodzik:
I added unit tests, but need the other PR to do some integration ones. Anyone willing to take a look? :sweat_smile:

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
```diff
@@ -0,0 +1,52 @@
...
+        }
+        val sourceJars = lib.sources.map { uri -> uri.toString() }.map {
+            val artifact = MavenDependencyModuleArtifact(it)
+            artifact.classifier = "sources"
+            artifact
+        }
+
+        // Matches the Maven group (organization), artifact, and version in the Bazel dependency
+        // string such as .../execroot/monorepo/bazel-out/k8-fastbuild/bin/external/maven/com/google/guava/guava/31.1-jre/processed_guava-31.1-jre.jar
+        val regexPattern = """.*/maven/(.+)/([^/]+)/([^/]+)/[^/]+.jar""".toRegex()
```
tgodzik, line 26: As far as I know there is no other way to get the coordinates, but the current regex seems to work and also the actual data is not used by Metals at all, just the relationship between source and normal jars.

Tested it on a large repo and it seems to work fine.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapper.kt
```diff
@@ -68,23 +75,25 @@ import org.jetbrains.bsp.bazel.server.sync.languages.jvm.javaModule
...
 import kotlin.io.path.relativeToOrNull
 import kotlin.io.path.toPath
 
 class BspProjectMapper(
-  private val languagePluginsService: LanguagePluginsService,
-  private val workspaceContextProvider: WorkspaceContextProvider,
-  private val bazelPathsResolver: BazelPathsResolver,
-  private val bazelRunner: BazelRunner,
-  private val bspInfo: BspInfo,
+    private val languagePluginsService: LanguagePluginsService,
```
agluszak, line 92: where do these formatting changes come from?

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
```diff
@@ -0,0 +1,52 @@
...
+        while (toResolve.isNotEmpty()){
+            val lib = project.libraries[toResolve.pop()]
+            if (lib != null && !accumulator.contains(lib)) {
+                accumulator.add(lib)
+                toResolve.addAll(lib.dependencies)
+            }
+        }
+        return accumulator
+    }
+}
```
agluszak, line 52: no newline

server/src/test/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapperTest.kt
```diff
@@ -0,0 +1,104 @@
...
+import java.net.URI
+import java.nio.file.Paths
+
+class DependencyMapperTest {
+
+    private val cacheLocation = "file:///home/user/.cache/bazel/_bazel_user/ae7b7b315151086e31e3b97f9ddba009/execroot/monorepo/bazel-out/k8-fastbuild-ST-4a519fd6d3e4"
+
+    @Test
+    fun `should translate dependency`() {
+        val jarUri = URI.create("$cacheLocation/bin/external/maven/org/scala-lang/scala-library/2.13.11/processed_scala-library-2.13.11.jar")
```
agluszak, line 22: does it work both with and without bzlmod?

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
agluszak: okay, I'm 99% sure it doesn't work with bzlmod.

Paths with bzlmod enabled will look something like `bazel-out/k8-fastbuild/bin/external/rules_jvm_external~override~maven~maven/commons-cli/commons-cli/1.6.0/commons-cli-1.6.0.jar`

agluszak (changes requested):

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
agluszak: So in case of bzlmod different versions of rules_jvm_external will use different paths, + changing the name of the repo generated by `maven.install` will also affect the path.

The only stable thing is `~maven~` part, which contains the name of the extension.

See `bazel-out/k8-fastbuild/bin/external/rules_jvm_external~~maven~name/v1/https/repo1.maven.org/maven2/com/google/auto/service/auto-service-annotations/1.1.1/header_auto-service-annotations-1.1.1.jar` for example

server/src/test/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapperTest.kt
tgodzik: I will check and fix it, thanks for taking a look!

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapper.kt
tgodzik: Fixed, Intellij didn't like the 8 space indents.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
tgodzik: Fixed

server/src/test/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapperTest.kt
tgodzik: Should work now.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
tgodzik: I changed the logic to use label to get the start of the organisation name and then the regex seems to work in both cases. Not perfect by any means, but I don't know if we can actually get organisation anyway else.

agluszak (approved):

tgodzik:
Thanks!

PR #544, opened 2024-03-19
bugfix: Fix showing Scala 3 diagnostics

tgodzik:
Previously, Scala 3 diagnostics would not be shown since they currently use rendered message, which is a pretty printed diagnostic. Now, I added a separate parser for Scala 3 compiler diagnostics.

Related https://github.com/scalameta/metals/issues/6135


agluszak (approved):

agluszak:
could you add an entry to the changelog? https://github.com/JetBrains/bazel-bsp/blob/master/CHANGELOG.md

tgodzik:
> could you add an entry to the changelog? https://github.com/JetBrains/bazel-bsp/blob/master/CHANGELOG.md

Done! Thanks

PR #559, opened 2024-04-23
fix: Fix issues with warnings being cleared too soon

tgodzik:
It seems that it was working at some point but somehow it stopped and I am not 100% sure why.

I noticed two issues currently:
- build target label is sometime missing @ so I add them to the label if we are currently in a target that has them
- the warnings would be cleared sine the build was successful, so remove files with warnings from being cleared.

I added a tests that tries to get the warnings twice in the same build whihc should test this properly.

### DIFF
--- a/.teamcity/configurations/bazelBsp/bazelE2eTests.kts
+++ b/.teamcity/configurations/bazelBsp/bazelE2eTests.kts
@@ -67,6 +67,7 @@ object AndroidProjectTest : BazelBspE2ETestsBuildType(
 object ScalaProjectTest : BazelBspE2ETestsBuildType(
     targets = "//e2e:enabled_rules_test",
 )
+
 object AndroidKotlinProjectTest : BazelBspE2ETestsBuildType(
     targets = "//e2e:android_kotlin_project_test",
 )
--- a/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.kt
+++ b/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.kt
@@ -284,7 +284,12 @@ class BepServer(
   }
 
   private fun consumeCompletedEvent(event: BuildEventStreamProtos.BuildEvent) {
-    val label = event.id.targetCompleted.label
+    val eventLabel = event.id.targetCompleted.label
+    /* The events never contain @, which will be different than the actual target id. Here we work around that fact,
+    * but since we also set up the BEP server to gather info about build targets within certain path (//... etc.), we can't
+    * just use target.uri. So this only fixes the diagnostics without breaking the query for targets.
+    * */
+    val label = if (target != null && ("@$eventLabel" == target.uri || "@@$eventLabel" == target.uri) ) target.uri else eventLabel
     val targetComplete = event.completed
     val outputGroups = targetComplete.outputGroupList
     LOGGER.trace("Consuming target completed event {}", targetComplete)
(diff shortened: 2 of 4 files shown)


.teamcity/configurations/bazelBsp/bazelE2eTests.kts
```diff
@@ -60,6 +60,10 @@ object KotlinProjectTest : BazelBspE2ETestsBuildType(
     targets = "//e2e:kotlin_project_test",
 )
 
+object ScalaProjectTest : BazelBspE2ETestsBuildType(
```
tgodzik, line 63: It seems this hasn't really added the test? Anyone has any idea what is going on?

Also how to debug the tests? Any tips?

hb-man: regarding debugging (if it doesn't work) @abrams27 may be of help

tgodzik: > yes, it will require adding test in on other place - i'll make a commit with required change to your branch tommorow.

Thanks!

hb-man:
I've added the test to TC pipeline, however, until it merged in master - it won't appear in the pipeline.
I've configured temporary test run on TC for this PR, so it will run on your changes and you can see how it behaves on TC.

tgodzik:
Thanks! It doesn't seem I can access it to see what is wrong. Will try to see if I can fix it locally.

hb-man:
> Thanks! It doesn't seem I can access it to see what is wrong. Will try to see if I can fix it locally.

should be available from now on, you can check again.
also, here are log files for failed tests:
[enabled_rules_test_bazel_5_3_2.txt](https://github.com/JetBrains/bazel-bsp/files/15094545/enabled_rules_test_bazel_5_3_2.txt)
[enabled_rules_test_bazel_6_4_0.txt](https://github.com/JetBrains/bazel-bsp/files/15094546/enabled_rules_test_bazel_6_4_0.txt)

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.kt
```diff
@@ -284,7 +284,8 @@ class BepServer(
   }
 
   private fun consumeCompletedEvent(event: BuildEventStreamProtos.BuildEvent) {
-    val label = event.id.targetCompleted.label
+    val eventLabel = event.id.targetCompleted.label
+    val label = if (target != null && ("@$eventLabel" == target.uri || "@@$eventLabel" == target.uri) ) target.uri else eventLabel
```
agluszak, line 288: can you explain the logic behind this?

tgodzik: It looks like sometimes the label is without `@` or `@@` so we don't identify the target properly. And for the on progress updates we sue the actual build target label

agluszak: hmm, do you have any idea why it sometimes contains `@` and sometimes not?

tgodzik: I think it might be a bug in Bazel, I think this is why I had to do similar fix before. I can dig in a bit more later.

tgodzik: Ach, I remember now, the events never contain `@` this is the exact same reason we had to do the workaround in https://github.com/JetBrains/bazel-bsp/pull/534/files

So here this works around that fact, but if we set up the BEP server to gather info about build targets within certain path (//... etc.) then this will not break. So this only fixes the diagnotics.

The problem happens because in progress notifications we don't get the target label, so we use whatever BEP was set up with.

agluszak: Can you put that as a comment in the code?

tgodzik: Done, this still isn't perfect, since if we retry a compilation on a target, which was already compiler with warnings they will be cleaned as the build is cached and no new events are produced. But this is a different issue not sure how to fix it properly yet.

tgodzik: Though I think it's still an improvement as before this PR warnings would just disappear after any changes, @agluszak does it make sense to merge?

agluszak (approved):

PR #542, opened 2024-03-14
feature: Add support for buildTarget/jvmCompileClasspath

tgodzik:
Previously, classpath would be obtained via scalacOptions and javacOptions requests, which would cause additional querries even if client was only interested in actual javac/scalac options.

Now, we don't send classpath together with scalac/javac options if the client declares that they will use the new jvmCompileClasspath request, which they should then trigger.

Integration tests blocked on https://github.com/build-server-protocol/bsp-testkit2/pull/43

PR #560, opened 2024-04-24
bugfix: Don't send classpath if jvmClasspathReceiver is set to true

tgodzik:
It seems that somehow I missed adding it in the previous PR, which baffles me greatly.

This makes sure that if the client wants to only receive classpath via jvmClasspath request then it's done properly.

I totally missed it in https://github.com/JetBrains/bazel-bsp/pull/542/files

### DIFF
--- a/e2e/src/main/kotlin/org/jetbrains/bsp/bazel/BazelBspScalaProjectTest.kt
+++ b/e2e/src/main/kotlin/org/jetbrains/bsp/bazel/BazelBspScalaProjectTest.kt
@@ -27,9 +27,9 @@ import kotlin.time.Duration.Companion.minutes
 import kotlin.time.Duration.Companion.seconds
 
 object BazelBspScalaProjectTest : BazelBspTestBaseScenario() {
-  private val testClient = createTestkitClient()
-  private val testClientClasspathReceiver = createTestkitClient(true)
   private val log = LogManager.getLogger(BazelBspScalaProjectTest::class.java)
+  private val testClient = createTestkitClient()
+  private val testClientClasspathReceiver = createTestkitClient(jvmClasspathReceiver = true)
 
   @JvmStatic
   fun main(args: Array<String>) = try{
--- a/e2e/src/main/kotlin/org/jetbrains/bsp/bazel/BazelBspScalaProjectTest.kt
+++ b/e2e/src/main/kotlin/org/jetbrains/bsp/bazel/BazelBspScalaProjectTest.kt
@@ -28,6 +28,7 @@ import kotlin.time.Duration.Companion.seconds
 
 object BazelBspScalaProjectTest : BazelBspTestBaseScenario() {
   private val testClient = createTestkitClient()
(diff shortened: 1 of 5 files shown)


server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectSyncService.kt
```diff
@@ -41,12 +42,17 @@ import org.jetbrains.bsp.WorkspaceDirectoriesResult
...
 import org.jetbrains.bsp.bazel.server.sync.model.Language
+import java.util.Optional
+import kotlin.jvm.optionals.getOrDefault
 
 /** A facade for all project sync related methods  */
 class ProjectSyncService(private val bspMapper: BspProjectMapper, private val projectProvider: ProjectProvider) {
 
-  fun initialize(cancelChecker: CancelChecker): InitializeBuildResult =
-    bspMapper.initializeServer(Language.all())
+  private var clientCapabilities = Optional.empty<BuildClientCapabilities>()
```
agluszak, line 51: You probably want to use a [lateinit var](https://kotlinlang.org/docs/properties.html#late-initialized-properties-and-variables) here

e2e/src/main/kotlin/org/jetbrains/bsp/bazel/BazelBspScalaProjectTest.kt
```diff
@@ -8,13 +8,15 @@ import kotlin.time.Duration.Companion.seconds
 
 object BazelBspScalaProjectTest : BazelBspTestBaseScenario() {
   private val testClient = createTestkitClient()
+  private val testClientClasspathReceiver = createTestkitClient(true)
```
agluszak, line 11:
```suggestion
  private val testClientClasspathReceiver = createTestkitClient(jvmClasspathReceiver = true)
```

agluszak (changes requested):

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ProjectSyncService.kt
```diff
@@ -41,12 +42,18 @@ import org.jetbrains.bsp.WorkspaceDirectoriesResult
 import org.jetbrains.bsp.WorkspaceInvalidTargetsResult
 import org.jetbrains.bsp.WorkspaceLibrariesResult
 import org.jetbrains.bsp.bazel.server.sync.model.Language
+import java.util.Optional
+import kotlin.jvm.optionals.getOrDefault
```
agluszak, line 46:
```suggestion
```
No longer needed I guess?

tgodzik:
I am fixing tests separately as they were never run on CI. We can wait until those are passing if you want?

tgodzik:
@agluszak I rebased on master and the new tests pass, do you think it's good to merge?

agluszak (approved):

PR #561, opened 2024-04-24
[feature] include tool tag for all Bazel invocations

mnoah1:
Add a global tool_tag for all Bazel invocations from Bazel BSP, which will include the server name and version.  In setups where Bazel telemetry is collected, this will allow Bazel BSP invocations to be easily tracked and segmented out from other Bazel calls.

https://youtrack.jetbrains.com/issue/BAZEL-932


agluszak:
LGTM. @abrams27?

agluszak (approved):

abrams27:
if the flag doesnt invalidate caches im good with it

abrams27:
yea but im referring to already existing caches, it's quite important for us

mnoah1:
Ok sounds good - yeah I was going to say that flag has always been part of the old [Bazel plugin](https://github.com/bazelbuild/intellij/blob/86e31292434875ca1f86a02230cde5a3de5a70cf/base/src/com/google/idea/blaze/base/command/BlazeFlags.java#L97) without issue, and the concern about it affecting performance when switching back and forth between plugin and cli was debunked here (https://github.com/bazelbuild/intellij/issues/789).  It's also categorized in bazel help under the flags that only affect logging.

PR #563, opened 2024-04-25
Support coverage execution in BuildTargetTest run

mnoah1:
Add a way to execute a test run with coverage for a single request:
- New BazelTestParamsData class to populate the data field of the TestRunParams.  This can be used to include other data (e.g. test filters, other flags, etc) that may be needed in the future for individual test runs.
- During the "test" run, substitute a "coverage" command to replace the Bazel test command.
- Get the coverage report from testActionOutputs and include in a progress event to the client.
  
https://youtrack.jetbrains.com/issue/BAZEL-930


mnoah1:
Hi @agluszak @abrams27 

While looking at this one I had a few questions:
- The test run here currently executes a build before running the test.  Is there a reason not to just let Bazel handle this as part of the test execution command?  As there are additional variations of testing introduced that could involve different builds (e.g. regular test, coverage, maybe test debugging in the future), it removes the complication of having to get the right combination of build flags to match up with the upcoming test run.
- Format to return the coverage results - how to distribute the work between server and client:
  - Server could parse the actual file and return an object containing the results.  Format for an object representing LCOV data would then likely need to be part of the protocol.  (client would probably still need to do some conversion of the returned result back to its own required format to display)
  - Server could just return the path to the coverage file and let the client do what it wants.

agluszak:
Thank you very much for this contribution!

> The test run here currently executes a build before running the test. Is there a reason not to just let Bazel handle this as part of the test execution command?

I don't think there's any specific reason for that. Probably that's a leftover from some legacy incarnation of the server :) Probably we wanted to make sure that all compilation notifications (especially diagnostics) come from the build request to make implementation of the client easier. But it should be possible to change that.

> Format to return the coverage result
I feel that the approach you chose is indeed better. Previously we've run into problems with either underdefining or overdefining stuff in the protocol, but the less data there is to be passed the lesser the problem ;)

mnoah1:
Ok, I removed the extra build step.  Do we have any BuildTestTarget related testing currently that should be updated to add this scenario?  Other than that this should be ready for review.

mnoah1:
@agluszak @abrams27 do you have any other feedback on this one?

agluszak:
We're both on vacation till next week :)

mnoah1:
Oh, ok, sorry about that! :)

bazelrunner/src/main/kotlin/org/jetbrains/bsp/bazel/bazelrunner/params/BazelFlag.kt
```diff
@@ -16,6 +16,9 @@ object BazelFlag {
   @JvmStatic fun buildManualTests(): String =
           flag("build_manual_tests")
 
+  @JvmStatic fun buildEventBinaryPathConversion(enabled: Boolean): String =
+    arg("build_event_binary_file_path_conversion", if (enabled) "true" else "false")
```
abrams27, line 20: why not just `enabled.toString`?

protocol/src/main/kotlin/org/jetbrains/bsp/TestCoverageReport.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.bsp
+
+public data class TestCoverageReport (
+        val lcovReportPath: String
```
abrams27, line 4: usually in the protocol we use uri instead of path

abrams27 (commented):
just small comments, but yea let's get back to it next week

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.kt
```diff
@@ -126,6 +126,11 @@ class BepServer(
         else -> TestStatus.FAILED
       }
 
+      val coverageReportUri = testResult.testActionOutputList.find { it.name == "test.lcov" }?.uri
+      if (coverageReportUri != null) {
+        bspClientTestNotifier.sendCoverageReport(parentId, coverageReportUri)
+      }
+
```
agluszak, line 133: the progress notification is going to be sent before the test even started ;)

logger/src/main/kotlin/org/jetbrains/bsp/bazel/logger/BspClientTestNotifier.kt
```diff
@@ -52,6 +54,20 @@ class BspClientTestNotifier(private val bspClient: BuildClient, private val orig
...
   }
 
+  /**
+   * Notifies the client of an available coverage report file.
+   *
+   * @param taskId         TaskId of the task that produced the coverage report
+   * @param lcovReportPath full path path to the coverage report file
+   */
+  fun sendCoverageReport(taskId: TaskId, lcovReportPath: String) {
+    val taskProgressParams = TaskProgressParams(taskId)
```
agluszak, line 64: hmm, I think it's a very good question whether this should be a task progress notification, a separate task or maybe a completely separate notification (like sending diagnostics). WDYT @jastice?

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ExecuteService.kt
```diff
@@ -77,20 +79,34 @@ class ExecuteService(
...
-        var result = build(cancelChecker, targets, params.originId)
-        if (result.isNotSuccess) {
-            return TestResult(result.statusCode)
-        }
         val targetsSpec = TargetsSpec(targets, emptyList())
 
+        var bazelTestParamsData: BazelTestParamsData? = null
+        try {
+            if (params.dataKind == BazelTestParamsData.DATA_KIND) {
+                val rawParams = params.data.toString()
```
agluszak, line 87: I don't know if that's going to work, in other places ([for example the client](https://github.com/JetBrains/intellij-bsp/blob/44cd7cdfdd2d9458dcde94ecfa4c375c47abaab6/src/main/kotlin/org/jetbrains/plugins/bsp/server/client/BspClient.kt#L86)) we use 

```
gson.fromJson(params.data as JsonObject, Class::class.java)
```

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ExecuteService.kt
```diff
@@ -77,20 +79,34 @@ class ExecuteService(
...
+                bazelTestParamsData = Gson().fromJson(rawParams, BazelTestParamsData::class.java)
```
agluszak, line 88: Gson doesn't have to be constructed every time

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ExecuteService.kt
```diff
@@ -77,20 +79,34 @@ class ExecuteService(
...
-                .withArguments(params.arguments)
-                .withFlag(BazelFlag.color(true))
-                .executeBazelBesCommand(params.originId, bepReader.eventFile.toPath())
-                .waitAndGetResult(cancelChecker, true)
+        val result = withBepServer(params.originId, params.targets.single()) { bepReader ->
+            run {
+                baseCommand
+                    .withTargets(targetsSpec)
+                    .withArguments(params.arguments)
+                    .withFlag(BazelFlag.buildEventBinaryPathConversion(false))
```
agluszak, line 105: why is that needed?

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/BepServer.kt
mnoah1: Since this report is associated with the target, I had put it as a progress notification on the target's test task (sent on line 107), instead of as progress on an individual test case .  Let me know what you think.

agluszak: ahh, yes, my bad. Then it looks good :)

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ExecuteService.kt
mnoah1: It provides file:// instead of bytestream:// paths in the outputs when using remote cache. ([--build_event_binary_file_path_conversion](https://bazel.build/reference/command-line-reference#flag--build_event_binary_file_path_conversion)).

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ExecuteService.kt
agluszak: or rather: if it's going to work, it will probably do a to string/from string roundtrip that could otherwise be avoided

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/ExecuteService.kt
agluszak: Could you put that explanation as a comment in the code?

mnoah1: Added

logger/src/main/kotlin/org/jetbrains/bsp/bazel/logger/BspClientTestNotifier.kt
mnoah1: Yeah good point.  My thought process in putting this under progress and not under its own task is that this this is one output file among potentially several others within the overall test run task, and not inherently a task on its own.

However that's a good point that we could potentially leverage diagnostics here:
- bazel-bsp could handle parsing of the coverage file
- conversion to document ranges (basically 1 diagnostic for every line of each file in the run)
- sending those to the client as diagnostics which include execution counts per line
- the client would then need to have some handling that treats these diagnostics as coverage highlighting instead of regular diagnostics

I think it might get a bit more involved, maybe we could start off with the initial step of exposing a path to the coverage file as it is now, and looking at diagnostics as a next step?

jastice: Agree, ideally we'd use diagnostics here, maybe with additional `CoverageDiagnosticData`. This could include the file as first step

It seems a bit off to use progress, unless there is actual progress to report?

mnoah1: Ok, I can try switching it over to a diagnostic.  Only thing is that publishDiagnostics doesn't seem great for sending a single file path either, as it requires a TextDocumentIdentifier in the notification, but the coverage report is a target level output that doesn't belong to a particular location in any one document.  Maybe an optional field on the test report could be another option.

mnoah1: For the first step of just returning the file, I tried to switch this over to diagnostics, but the part about assigning it to a TextDocumentIdentifier feels a bit forced, since it's more of a top-level output of the test target and not associated with any one text document.  Also, even if we do pick a document to add there, the user would see a uri to a file that is not really human readable and is meant for processing by the client.

The _contents_ could potentially be parsed and converted to something diagnostic-like, but earlier in the thread we had discussed that at least for now, the client can parse it.

It feels like data that should be more associated with a task than with a document, either as its own task, progress on the target task (even though it's not numerical progress, "this new output is available" is still a form of progress), or another field in the test finish data.  Thoughts on any of those?

agluszak: What I meant with diagnostics was not necessarily reusing the existing diagnostics notification, but rather making a new one, completely separate from the task notification system ;)

mnoah1: Oh ok, that makes more sense :)  So maybe a new notification called something like `build/publishOutput` and params like this?  It can be used for similar cases where there is a need to send an output from the build to the client.

```
export interface PublishOutputParams {
  /** The build target where the output originates.
  buildTarget: BuildTargetIdentifier;

  /** The request id that originated this notification. */
  originId?: OriginId;

  /** Kind of data to be included here. */
  dataKind: string,

  /** Data matching dataKind above. */
  data: Any;
}
```

jastice: That sounds reasonable, if we have other good use cases for such a notification (test report files? build artifacts?). Otherwise I might just go with a dedicated one for this use-case.

mnoah1: >if we have other good use cases for such a notification (test report files? build artifacts?)

I don't have any immediately off the top of my head, but yeah I was thinking other cases like those might come up in the future and it would be easier to support them if the notification is more general.

mnoah1: I made the update above.  For the Client, it looks like this is the first notification type outside the existing protocol, so I had to create a JoinedBuildClient, similar in approach to the JoinedBuildServer - as I had to update this in a few other places, let me know if I did anything incorrect there.

I made the params essentially the same as above, and notification `build/publishOutput`.  I was able to test this with our client and received the notification.

Let me know if you all have any other feedback on this.

mnoah1:
@agluszak @abrams27 let me know if you have any other feedback prior to this being ready to merge

agluszak (approved):
Just nits :)

agluszak:
Rebase please

mnoah1:
Ok, updated this, it should be able to merge now.

PR #568, opened 2024-05-07
Parse test.xml results following a test run

mnoah1:
Set up parsing of test xml results following a test run:
- After a test run, check the testActionOutputs for presence of "test.xml" file, which is a standard output from most Bazel test actions and provides results in JUnit format across languages.
- If present, deserialize the xml contents, and convert into task notifications, using parent-child relationship to indicate nesting in order of target -> suite -> testCase
- New TestCaseTestFinishData/TestSuiteFinishData which can be used to populate additional details
- Since test.xml is standard across test actions, getting this working would at least ensure that there's a way to provide test case level pass-fail info at the end of a run without additional work needed per language.  This could supplement future work to watch the output as it occurs and provide more test framework-specific parsing that provides live results.

https://youtrack.jetbrains.com/issue/BAZEL-614


mnoah1:
As we had discussed during our call, I began taking a look at converting the test.xml results into notifications.  This isn't quite done yet but I wanted to get alignment on a few points:
- In the discussion (https://github.com/build-server-protocol/build-server-protocol/discussions/652), I want to confirm how to represent groups of tests.
  - Are we finalized on using a separate type, e.g. "test-group-start" to represent the group of tests? If so then I can use that for the target as well as the suite (multiple groups nested under each other).
  - For the TestFinishData, I defined TestCaseTestFinishData and TestSuiteTestFinish data to contain more specifics.  These would contain more specifics about the error message - alternatively we could publish diagnostics, maybe both?  I'm thinking of cases where we want to provide the user with clear stderr/stdout related to a given test case, that may not easily tie to a specific range in a file.
- Since test.xml is pretty universally available, but also not available till the end of a target's run, I figured this can be a starting point, and more real-time watchers/parsers could be added to provide more real-time data for specific test frameworks as needed with this becoming a fallback.

cc @agluszak @abrams27

jastice:
> Are we finalized on using a separate type, e.g. "test-group-start" to represent the group of tests? If so then I can use that for the target as well as the suite (multiple groups nested under each other).

Do we need separate concepts for groups and groups-of-groups? Not all test frameworks may differentiate like that, and they might have multiple layers of nesting. Or am I misunderstanding the idea?

>Since test.xml is pretty universally available, but also not available till the end of a target's run, I figured this can be a starting point, and more real-time watchers/parsers could be added to provide more real-time data for specific test frameworks as needed with this becoming a fallback.

sounds good to me

mnoah1:
>Do we need separate concepts for groups and groups-of-groups? Not all test frameworks may differentiate like that, and they might have multiple layers of nesting. Or am I misunderstanding the idea?

Yeah, it seems like the concept of a suite/test group is sufficiently implied if a test item has multiple children, and that giving it a separate type it might not be totally necessary and might even become a bit restrictive.  Commented on the discussion here https://github.com/build-server-protocol/build-server-protocol/discussions/652

mnoah1:
I went ahead and updated this, using a single type to hold whatever details are available to be parsed from that entry in the test results, and the parent-child relationship derived from the parents in the TaskId.  Please review and let me know if you have feedback.

server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/TestXmlParser.kt
```diff
@@ -0,0 +1,194 @@
...
+        val errorType = when {
+            testCase.error != null -> testCase.error.type
+            testCase.failure != null -> testCase.failure.type
+            else -> null
+        }
+        val testCaseData = TestCaseTestFinishData(testCase.time, testCase.classname, parentSuite.pkg, fullOutput, errorType)
+        bspClientTestNotifier.startTest(testCase.name, testCaseTaskId)
+        bspClientTestNotifier.finishTest(testCase.name, testCaseTaskId, testStatusOutcome, outcomeMessage, TestCaseTestFinishData.DATA_KIND, testCaseData)
+    }
+}
```
agluszak, line 194:
```suggestion
}

```

server/src/test/kotlin/org/jetbrains/bsp/bazel/server/bep/TestXmlParserTest.kt
```diff
@@ -0,0 +1,370 @@
...
+            }
+        }
+    }
+
+    private fun writeTempFile(tempDir: Path, contents: String): String {
+        val tempFile = tempDir.resolve("tempFile.xml").toFile()
+        tempFile.writeText(contents)
+        return tempFile.toURI().toString()
+    }
+}
```
agluszak, line 370:
```suggestion
}

```

agluszak (commented):
Again, just nits :) looks very good!

protocol/src/main/kotlin/org/jetbrains/bsp/TestCaseTestFinishData.kt
```diff
@@ -0,0 +1,13 @@
+package org.jetbrains.bsp
+
+public data class TestCaseTestFinishData (
+        val time: Double,
+        val className: String?,
+        val pkg: String?,
+        val fullError: String?,
+        val errorType: String?,
+) {
```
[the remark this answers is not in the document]
mnoah1: The format of the XML file originated from JUnit, but now has become a de-facto standard more widely.  For other languages, the report generators generally seem to try to populate these fields with the closest equivalents in the corresponding language (e.g. classname), or leave it off (hence leaving most fields nullable here so we can just capture as much as is available but not depend on it to all be present) - but still adhere to the same overall structure.  I guess I can do `JUnitStyleTestCaseData` to make it a little clearer that the tests results are not necessarily JUnit, but transformed into this format.

Go example with rules_go:
```
<testsuite errors="0" failures="0" skipped="0" tests="11" time="0.017" name="sample/path/to/packagename">
		<testcase classname="packagename" name="TestSampleTestCase" time="0.000"></testcase>
				  ^ reused for the go package name
</testsuite>
```

Python example with rules_python:
```
 <testsuite name="pytest" errors="0" failures="0" skipped="0" tests="5" time="0.412" timestamp="2024-05-20T17:48:23.933620" hostname="sample">
        <testcase classname="example.path.to.tests.test_file" name="test_case_1" time="0.001" />
                                               ^ for a plain test function directly in a file
        <testcase classname="example.path.to.tests.test_file.TestClass" name="test_case_2" time="0.001" />
                                               ^ for a test class
</testsuite>
```

agluszak (approved):

PR #592, opened 2024-06-14
[fix] Allow arbitrary ending to Scala 3 error line

tgodzik:
It turns out the ---- ending is not always showing up (or it changed) so improved the regex to account for that.


tgodzik:
Anyone could take a look? This is really a small change

xuansontrinh:
LGTM

xuansontrinh:
@tgodzik may I merge it?

